PR #193, opened 2014-08-31
Implement state provider for OAuth2 protocol

akkie:
Currently, to prevent the client for CSRF attacks as described in the [OAuth2 RFC](https://tools.ietf.org/html/rfc6749#section-10.12), the OAuth2 provider stores the state on the application side in the cache. So when using a clustered environment a distributed cache instance must be used. This pull request removes this cache dependency in favor of an abstract solution to handle the state on applications side. It comes with a session state provider that can store the state in the session to follow Play's stateless approach.


hellboy81:
why are I saving oauth2 state in `cookie`, but not in `session`?

akkie:
The Play session is also saved in a cookie. And by using an extra cookie we do not need to implement an expire strategy, because a cookie has such a strategy already implemented?

PR #233, opened 2014-12-09
make getAccessToken(code: String) accessible

datalek:
Making the method public allows us to follow the workflow described in [this](https://aaronparecki.com/articles/2012/07/29/1/oauth2-simplified#web-server-apps) link.

This workflow is used in some angular library like [this](https://github.com/sahat/satellizer/wiki/Login-with-OAuth-2.0)

### DIFF
--- a/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala
+++ b/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala
@@ -134,7 +134,7 @@ abstract class OAuth2Provider(httpLayer: HTTPLayer, stateProvider: OAuth2StatePr
    * @param code The access code.
    * @return The info containing the access token.
    */
-  protected def getAccessToken(code: String): Future[OAuth2Info] = {
+  def getAccessToken(code: String): Future[OAuth2Info] = {
     httpLayer.url(settings.accessTokenURL).withHeaders(headers: _*).post(Map(
       ClientID -> Seq(settings.clientID),
       ClientSecret -> Seq(settings.clientSecret),


akkie:
Do you have found a solution for the state param. I try to implement the Angular Seed Template with Satellizer and now I have the same problem with the state provider. It seems that Satellizer doesn't support the state param at all.

merle-:
No Christian, now i'm using a dummy state that always validate the state
Il 04/gen/2015 14:47 "Christian Kaps" <notifications@github.com> ha scritto:

Do you have found a solution for the state param. I try to implement the
Angular Seed Template with Satellizer and now I have the same problem with
the state provider. It seems that Satellizer doesn't support the state
param at all.

—
Reply to this email directly or view it on GitHub
<https://github.com/mohiva/play-silhouette/pull/233#issuecomment-68633425>.

akkie:
@merle- 

I've created a new [issue](https://github.com/sahat/satellizer/issues/238) to support the state parameter.

akkie:
@merle- What do you think if we make the state provider optional? @sahat raised a valid [statement](https://github.com/sahat/satellizer/issues/238#issuecomment-68772829) about the client side state validation.

merle-:
Yes, in my opinion if the state provider become optional play-silhouette will be more felxible on workflow to use

akkie:
What do you think is the best solution?

- Provide a (Null|Dummy)StateProvider
- Make the state optional

If we make the state optional then the code complexity in the authenticate method increases. We break also the API for projects using a OAuth2 provider.

merle-:
In my opinion provide a `(Null, Dummy, None)StateProvider` is not so bad and the code remain as it is, elegant and concise. To make it more transparent we can add to companion object of providers an apply method that set `DummyStateProvider` as default.
```scala
/**
 * The companion object.
 */
object FacebookProvider {

....

  /**
   * Creates an instance of the provider.
   *
   * @param httpLayer The HTTP layer implementation.
   * @param stateProvider The state provider implementation.
   * @param settings The provider settings.
   * @return An instance of this provider.
   */
  def apply(httpLayer: HTTPLayer, stateProvider: OAuth2StateProvider, settings: OAuth2Settings) = {
    new FacebookProvider(httpLayer, stateProvider, settings) with CommonSocialProfileBuilder
  }

  /**
   * Creates an instance of the provider.
   *
   * @param httpLayer The HTTP layer implementation.
   * @param settings The provider settings.
   * @return An instance of this provider.
   */
  def apply(httpLayer: HTTPLayer, settings: OAuth2Settings) = {
    new FacebookProvider(httpLayer, DummyStateProvider, settings) with CommonSocialProfileBuilder
  }
}
```

akkie:
I've added the `DummyState`. I've omitted the additional apply method, because this would introduce a lot of duplicate code which must also be tested. A better solution would be to put the code into a trait. But I had no right idea for a name at the moment.

merle-:
Yes you are right, I hadn't thought of that :/.

PR #240, opened 2014-12-19
[WIP] Akka HTTP

rfranco:
:exclamation: *WORKING IN PROGRESS* :exclamation:

the idea is split silhouette in multi project as core, play, play-testkit, akka-http and akka-http-testkit, so it can be use in different web frameworks.

Abstract Layer:
- [ ] RequestHeader
- [ ] Headers
- [ ] Result
- [ ] Lang
- [ ] WSRequestHolder
- [ ] Logger
- [ ] Json.format

any help always is welcome.


rfranco:
I did some research and that will be a difficult task because will change a lot of the silhouette core.
So maybe in the future i do.

akkie:
As said previously, generally I really like the idea. But I also see the great effort to put into this change. Maybe later ...

PR #251, opened 2015-01-08
Incoherence in time units

ulric260:
For BearerTokenAuthenticator and JWTAuthenticator there were some incoherences in time units between code and documentation.
According to the doc :
 - authenticatorIdleTimeout should be in seconds
 - authenticatorExpiry should be in minutes
Whereas it's was the opposite in the code (expiry in sec and idlTimeout in min).

Now both are in seconds.


akkie:
Good catch! Could you also update the `SessionAuthenticator` and the `CookieAuthenticator`. They have the same problem.

akkie:
Thanks, looks good. Please could you combine the two commits into one commit. You could do that with git interactive rebase. 

@see 
https://github.com/mohiva/play-silhouette/pull/221#issuecomment-61687447
http://www.openshift.org/documentation-latest/oo_contributors_guide.html#combining-multiple-commits

Thanks

PR #263, opened 2015-01-22
Update actions.rst

JamesSullivan:
You definitely want to look this over as I really do not know what I am doing with type parametric polymorphism in Scala. However, I absolutely could not get the WithProvider example to compile without those changes.

### DIFF
--- a/docs/how-it-works/actions.rst
+++ b/docs/how-it-works/actions.rst
@@ -196,8 +196,8 @@ logged in using a given provider:
 
 .. code-block:: scala
 
-  case class WithProvider(provider: String) extends Authorization[User] {
-    def isAuthorized(user: User)(implicit request: RequestHeader, lang: Lang) = {
+  case class WithProvider[I <: com.mohiva.play.silhouette.api.Identity](provider: String) extends Authorization[I] {
+    def isAuthorized(identity: I)(implicit request: RequestHeader, lang: Lang) = {
       user.identityId.providerId == provider
     }
   }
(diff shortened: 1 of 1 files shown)


akkie:
This compiles fine to me:

```scala
import com.mohiva.play.silhouette.api.{Authorization, LoginInfo, Identity}
import play.api.i18n.Lang
import play.api.mvc.RequestHeader

case class User(loginInfo: LoginInfo) extends Identity

case class WithProvider(provider: String) extends Authorization[User] {
  def isAuthorized(user: User)(implicit request: RequestHeader, lang: Lang) = {
    user.loginInfo.providerID == provider
  }
}
```

This is also tested with Scala 2.10.4 and Scala 2.11.4:

https://github.com/mohiva/play-silhouette/blob/master/silhouette/test/com/mohiva/play/silhouette/api/SilhouetteSpec.scala#L1045
https://github.com/mohiva/play-silhouette/blob/master/silhouette/test/com/mohiva/play/silhouette/api/SilhouetteSpec.scala#L138
https://github.com/mohiva/play-silhouette/blob/master/silhouette/test/com/mohiva/play/silhouette/api/SilhouetteSpec.scala#L157 

I've no idea what's going on in your code. Could you share a code snippet?

JamesSullivan:
Thanks for the quick reply Christian.

Looked into this a bit more and now understand that I need to import the
correct User.

import com.mohiva.play.silhouette.api.Authorization
import com.mohiva.play.silhouette.api.Identity
import play.api.i18n.Lang
import play.api.mvc.RequestHeader
// import com.mohiva.play.silhouette.impl.User  -- using this import causes
need to add type info
import models.User // this works without adding type info


case class WithProvider(provider: String) extends Authorization[User] {
  def isAuthorized(user: User)(implicit request: RequestHeader, lang: Lang)
= {
    // changed line below from example to work with version 2.0-SNAPSHOT
    // user.identityId.providerId == provider
    user.loginInfo.providerID == provider
  }
}

akkie:
Nice, that it works. Let me know if you have further questions.

PR #271, opened 2015-01-26
Change header in default Fingerprint generator

centaure:
In default FingerprintGenerator, replace ACCEPT header with ACCEPT-LANGUAGE header (when generating Fingerprint):

Using the ACCEPT header field for the default Fingerprint generator will cause problems for applications that use different media types depending on requests. For example, you may have a controller that processes user sign-in POST request RESTfully (accepting "application/json, text/javascript, /; q=0.01"); therefore the authenticator and its corresponding fingerprint is created using this JSON etc ACCEPT header. On the other hand, the post-sign-in process may include a Redirect action whose GET request uses the usual HTTP request type i.e. "text/html,application/xhtml+xml,application/xml;q=0.9". In this case, silhouette will fail during authenticator retrieval when processing the GET request (because the fingerprints are different). 

Although the user can implement his/her own generator in this case, it will be good to avoid the "ACCEPT" header for the default generator (so as to remove a potential complication for new users who, for example, use such a mixture of REST and HTML in their API). 

ACCEPT-LANGUAGE, on the other hand, will be least likely to be different across request types.

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/impl/util/DefaultFingerprintGenerator.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/util/DefaultFingerprintGenerator.scala
@@ -41,7 +41,7 @@ class DefaultFingerprintGenerator(includeRemoteAddress: Boolean = false) extends
   def generate(implicit request: RequestHeader) = {
     Crypt.sha1(new StringBuilder()
       .append(request.headers.get(USER_AGENT).getOrElse("")).append(":")
-      .append(request.headers.get(ACCEPT).getOrElse("")).append(":")
+      .append(request.headers.get(ACCEPT_LANGUAGE).getOrElse("")).append(":")
       .append(if (includeRemoteAddress) request.remoteAddress else "")
       .toString()
     )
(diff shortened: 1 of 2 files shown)


akkie:
The problem here is that the `Accept` header for a JSON request is defined by the JS library and not by the browser. So if you call a JSON endpoint from JS then you send another header as in a GET request initiated from your browser.

Let me think a bit about a solution. I'm not sure if we should change the default fingerprint generator. Maybe a better solution would be to create a new one for such a content negotiation scenario. Anyway, the `Accept-Language` header wouldn't be enough. The `Accept-Charset` and `Accept-Encoding` headers are also good candidates to generate a fingerprint of the client.

centaure:
I agree that "Accept-Charset and Accept-Encoding headers are also good candidates". 

FYI: I migrated to Silhouette v2.0-SNAPSHOT (from v1.0) in the last few days (using your silhouette-seed example as a guide). The above fingerprint issue was the only major issue I faced. Initially I had no clue what was going on (as I had not defined   "com.mohiva.play.silhouette.impl.authenticators.CookieAuthenticatorService" in my logger.xml file; CookieAuthenticatorService does print a logger info message when the fingerprints are different). But I figured it out after looking through the source code.

akkie:
I close this in favor of #276.

Thanks, for reporting the issue.

PR #277, opened 2015-01-29
Silhouette stopped working

akkie:
I've tried https://github.com/mohiva/play-silhouette-seed, for example. One, maybe two days ago it worked, now it doesn't. On my machine(Ubuntu 14.04) it works correctly only on Firefox, and not on Chrome, Opera and Vivaldi. The error is - after signing up I cannot sign in with email and password. authenticate action works correctly, but request.identity is None.


akkie:
I've published a new version with a fix. Thanks for reporting the issue.

malchmih:
Thank you for the library, I'm really enjoying it!

akkie:
You're welcome!

PR #279, opened 2015-02-01
Update Play 2.4.0

rfranco:
WIP

### DIFF
--- a/silhouette/test/com/mohiva/play/silhouette/api/util/DefaultEndpointHandlerSpec.scala
+++ b/silhouette/test/com/mohiva/play/silhouette/api/util/DefaultEndpointHandlerSpec.scala
@@ -2,6 +2,7 @@ package com.mohiva.play.silhouette.api.util
 
 import play.api.http.ContentTypes._
 import play.api.i18n.Messages
+import play.api.i18n.Messages.Implicits._
 import play.api.mvc._
 import play.api.test._
 
@@ -142,7 +143,7 @@ class DefaultEndpointHandlerSpec extends PlaySpecification {
     expectedContentType: String,
     expectedResponseFragment: String,
     expectedMessage: String,
-    f: RequestHeader => Future[Result]) = {
+    f: RequestHeader => Future[Result])(implicit messages: Messages) = {
     implicit val request = acceptedMediaType match {
       case Some(mediaType) => FakeRequest().withHeaders(ACCEPT -> mediaType)
       case None => FakeRequest()
--- a/silhouette/test/com/mohiva/play/silhouette/impl/authenticators/JWTAuthenticatorSpec.scala
+++ b/silhouette/test/com/mohiva/play/silhouette/impl/authenticators/JWTAuthenticatorSpec.scala
@@ -21,6 +21,7 @@ import com.mohiva.play.silhouette.api.LoginInfo
 import com.mohiva.play.silhouette.api.exceptions._
 import com.mohiva.play.silhouette.api.services.AuthenticatorService._
 import com.mohiva.play.silhouette.api.util.{ Base64, Clock, IDGenerator }
+import com.mohiva.play.silhouette.impl.authenticators.JWTAuthenticator._
 import com.mohiva.play.silhouette.impl.authenticators.JWTAuthenticatorService._
 import com.mohiva.play.silhouette.impl.daos.AuthenticatorDAO
 import org.joda.time.DateTime
@@ -30,7 +31,7 @@ import org.specs2.specification.Scope
 import play.api.libs.Crypto
 import play.api.libs.json.{ JsNull, Json }
 import play.api.mvc.Results
-import play.api.test.{ WithApplication, FakeRequest, PlaySpecification }
+import play.api.test.{ FakeRequest, PlaySpecification, WithApplication }
 
 import scala.concurrent.Future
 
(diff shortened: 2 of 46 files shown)


akkie:
Nice to see that you work on the Play 2.4 support!

rfranco:
I'm having problem with the new `MessageApi` that depends on `MessagesApi`.
Maybe need to refactor everything to not use global state.

@akkie Any idea?

akkie:
What do you mean with `MessagesApi`?
Do we use global state?

rfranco:
Play 2.4 has a new `MessagesApi` for i18n
https://www.playframework.com/documentation/2.4.x/ScalaI18N
and now `Messages` need a implicit `MessagesApi`.

There are some points which has problem as `DefaultEndpointHandler` and it's object.
That a look [here](https://github.com/rfranco/play-silhouette/blob/feature/play-2.4/silhouette/app/com/mohiva/play/silhouette/api/util/DefaultEndpointHandler.scala#L12) and [here](https://github.com/rfranco/play-silhouette/blob/feature/play-2.4/silhouette/app/com/mohiva/play/silhouette/api/util/DefaultEndpointHandler.scala#L49)

Other point i didn't figure out how to test the new `Messages`

akkie:
I think this is a bug in the `DefaultEndpointHandler`, because if a user overrides the default method (Accept-Language header) to retrieve the language, then this has no affect in the `DefaultEndpointHandler`. A better implementation would be to pass an implicit [Messages](https://github.com/playframework/playframework/blob/a0a8bd5796c746334dac1d88f65cda66975cb6a8/framework/src/play/src/main/scala/play/api/i18n/Messages.scala#L336) to `handleNotAuthenticated`, `handleNotAuthorized` and `produceResponse` methods.

What do you think?

rfranco:
@akkie finally i got time to finish it. Could you review pls

silhouette/test/com/mohiva/play/silhouette/api/util/DefaultEndpointHandlerSpec.scala
```diff
@@ -142,7 +143,8 @@ class DefaultEndpointHandlerSpec extends PlaySpecification {
     expectedContentType: String,
     expectedResponseFragment: String,
     expectedMessage: String,
-    f: RequestHeader => Future[Result]) = {
+    f: RequestHeader => Future[Result])(
+      implicit messages: Messages) = {
```
akkie: I think it's clearer to write it in one line.

```scala
    ...
    expectedMessage: String,
    f: RequestHeader => Future[Result])(implicit messages: Messages) = {

    implicit val request = acceptedMediaType match {
      case Some(mediaType) => FakeRequest().withHeaders(ACCEPT -> mediaType)
      case None => FakeRequest()
    }
```

silhouette/test/com/mohiva/play/silhouette/impl/authenticators/JWTAuthenticatorSpec.scala
```diff
@@ -63,7 +63,7 @@ class JWTAuthenticatorSpec extends PlaySpecification with Mockito with JsonMatch
       val jwt = service(None).serialize(authenticator)
       val json = Base64.decode(jwt.split('.').apply(1))
 
-      json must /("exp" -> authenticator.expirationDate.getMillis / 1000)
+      json must /("exp" -> (authenticator.expirationDate.getMillis / 1000).toInt) //FIXME There is not a implicit from Long to JsonSelect in Specs2
```
akkie: Why is this necessary now?

akkie:
Looks good, but there are a lot of build errors.

silhouette/test/com/mohiva/play/silhouette/impl/authenticators/JWTAuthenticatorSpec.scala
rfranco: Spec2 3.4 there isn't a implicit to convert `Long` to `JsonSelector`

https://github.com/etorreborre/specs2/blob/master/matcher-extra/src/main/scala/org/specs2/matcher/JsonMatchers.scala#L339-L344

rfranco:
@akkie for some reason the `Play Crypto.encryptAES` is return different value.

```scala
val secret = new CookieSecret("value", new DateTime(1429928030383))
secret.serialize // 2-S0NkeKD9j4buhu3ULMhTg8BIGiX5CCpzJKRkHEofNYRxg3OeCLku+62z1dt399NLmf9I7gu2SlcBDaDqUBqFRA==
secret.serialize // 2-ED7Baq3Ue1ktSq2yjqA5ORuPUzWoX/t8NlvYeDCCdD87XNZcJDLgKZghR3QvGj1Ym3My3C/JI42OVj5ETkoKWw==
```

I dont know why that happy, i think it's Play bug, Could you help to figure out how to solve?

akkie:
@rfranco The crypto API has changed in Play 2.4. The new API uses now `AES/CTR/NoPadding` with an initialization vector. Therefore it produces different values for the same input.

https://www.playframework.com/documentation/2.4.x/Migration24

rfranco:
@akkie Oh i can see so how could be solve that issue?

akkie:
@rfranco You have a pull request

rfranco:
@akkie oh you are the best! now look good to go! thanks

rfranco:
@akkie How should be the version now?

akkie:
I think we should pump the version to 3.0 because of the API changes which come with #326,  #296 and the execution context issue. Could you squash the commits into one and use the commit message "Update to Play 2.4"?

rfranco:
Silhuette need to give support for both Play 2.3.x and 2.4.x, my sugestion
is to do as Spray does to follow Akka version.

akkie:
With version 2.0 we have support for Play 2.3. I've not the time to maintain two not compatible versions. Silhouette uses [Semantic Versioning](http://silhouette.mohiva.com/v2.0/docs/releases#notes) and I think this is good because we can have our own lifecycle. Since the Play 2.3 release we had two major versions.

Play itself removes plugins from it's core to have a lifecycle independent of Play’s own lifecycle.

rfranco:
Ok, I agree, anyway the version's 3.0.0-SNAPSHOT

akkie:
@rfranco Thanks for your work on that. I really appreciate that.

PR #280, opened 2015-02-03
Remove LoginInfo from Identity trait

akkie:
@see https://groups.google.com/forum/#!topic/play-silhouette/XiItX-P3ZP0

### DIFF
--- a/docs/how-it-works/identity.rst
+++ b/docs/how-it-works/identity.rst
@@ -4,14 +4,26 @@ Identity
 ========
 
 Silhouette defines a user through its ``Identity`` trait. This trait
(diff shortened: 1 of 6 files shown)


guersam:
I like this change :) How about introducing `type FakeIdentity.Predicate` to make it a bit more explicit?

akkie:
I think we can remove the predicate function, if we use a Map[LoginInfo, Identity] for the FakeIdentityService and the FakeEnvironment.

guersam:
That makes sense. Just keeping it clean must be enough as it's not a part of user facing API :)

PR #284, opened 2015-02-14
CredentialsProvider & BasicAuthProvider should throw AccessDeniedExceptions

akkie:
Unless I'm misunderstanding https://github.com/mohiva/play-silhouette/blob/ce877f/docs/how-it-works/error-handling.rst
> The AccessDeniedException will be thrown if a user provides incorrect credentials

The following lines should throw AccessDeniedExceptions instead of AuthenticationExceptions:

[CredentialsProvider.scala#L76](https://github.com/mohiva/play-silhouette/blob/ce877f/silhouette/app/com/mohiva/play/silhouette/impl/providers/CredentialsProvider.scala#L76)

[BasicAuthProvider.scala#L77](https://github.com/mohiva/play-silhouette/blob/ce877f/silhouette/app/com/mohiva/play/silhouette/impl/providers/BasicAuthProvider.scala#L77)

### DIFF
--- a/docs/how-it-works/error-handling.rst
+++ b/docs/how-it-works/error-handling.rst
@@ -3,9 +3,9 @@ Error handling
 
 Every provider implementation in Silhouette throws either an
 ``AccessDeniedException`` or an ``AuthenticationException``. The
-``AccessDeniedException`` will be thrown if a user provides incorrect
-credentials or a social provider denies an authentication attempt. In
-any other case the ``AuthenticationException`` will be thrown.
+``AccessDeniedException`` will be thrown if a social provider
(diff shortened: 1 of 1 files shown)


akkie:
I'm currently search a good solution to fix issue https://github.com/mohiva/play-silhouette/issues/299 and currently I think your proposal of a third exception type could be a viable solution to fix this issue.

There could be a `NotAuthenticatedException` or `UnauthorizedException` which will be mapped to a 401 response. Then a `NotAuthorizedException` or a `ForbiddenException` which will be mapped to a 403 response. And at last an `AuthenticationException` for all other cases. But what will be done with this type of exception? Should the user handle this type of exception or should it also be mapped to a 401 response?

PR #289, opened 2015-02-20
Update REST seed with Silhouette 2.0

studiodev:
Hello,

I've updated the Silhouette-rest-seed to works with Silhouette 2.0.
I've tried to follow the way you develop the official seed.

The main difference is that I've removed the social things. I don't see in a REST API why we would like to connect with a social provider.
But If someone think it's important, I will accept PR gladly.

If you have any comment about the seed, just tell me, I can update it :)

### DIFF
--- a/docs/basics/examples.rst
+++ b/docs/basics/examples.rst
@@ -49,7 +49,7 @@ for dependency injection.
 Seed project for Play Framework with Silhouette, exposing a rest api for
 signup and signin.
 
-`Link <https://github.com/merle-/silhouette-rest-seed>`__
+`Link <https://github.com/studiodev/silhouette-rest-seed>`__
 
 --------------
 


akkie:
@studiodev This template wasn't developed by me but rather by @merle-. And I'm not sure if I should really replace this example with yours without asking @merle-. I hope you can understand this. So @merle- would it be OK to replace your example? We can also add a new example.

studiodev:
Yes of course I understand.

If @merle- want to continue to maintain this seed, I would gladly sent a PR to this project instead of linking my repository.

merle-:
Thanks @akkie and @studiodev, I've planned to update the example within this weekend, i'm going to use Cake Pattern for DI so i think that we could have two different example, one that use Guice for Dependency Injection and one that use Cake Pattern. Also i would like to cover link feature with other social profile and add a mail service.

studiodev:
Ok, sorry for the misunderstanding.

@akkie I can rename my seed "silhouette-minimal-rest-seed" so that you could publish it in addition of merle one, which will follow another approach.

Is it ok for you two?

akkie:
Sure, why not!

studiodev:
I will do another one

PR #294, opened 2015-02-26
Make FakeJWTAuthenticatorService working without started app

studiodev:
When the `encryptSubject` configuration is activated, the encryption rely on `Crypto.encryptAES` which require a started application (https://github.com/playframework/playframework/blob/2.3.x/framework/src/play/src/main/scala/play/api/libs/Crypto.scala#L59).

By just setting encryptSubject to false in test, it should allows to initiate the JWSAuthenticator before the WithApplication block.


akkie:
Nice catch. Thanks

Issue #242, opened 2014-12-26
CredentialsProvider doesn't play well with default Silhouette.exceptionHandler


Firfi:
Thing is on nonexist user it throws AccessDeniedException 

  new AccessDeniedException(CredentialsProvider.UnknownCredentials.format(id))

which exceptionHandler converts into 403 Forbidden status later in the flow that should be actually 401 as there's _no_ user which we deny access yet.

Controller code looks somewhat like this:

  ... case Some(p: CredentialsProvider) =>  p.authenticate(credentials) ...
  .flatMap(blahblah)
  ).recoverWith(exceptionHandler)

Firfi:
@akkie thanks for the answer. Actually I don't need fix immediately, but will send pull request anyway, just because I can. I found nothing better though than just change AccessDenied stuff to AuthenticationException. By the way, why use exceptions at all? They're for pussies. Maybe it'd be better to use some Try or Either?

akkie:
Try is already used inside Futures to handle errors. And this error handling works with exceptions. So there is no need for another error handling mechanism.

Thanks for the pull request

PR #298, opened 2015-03-03
changing log level from error to info when the user fails to type the co...

lkt:
...rrect credentials


akkie:
That's not so easy, because now all errors will be logged as info. With https://github.com/mohiva/play-silhouette/issues/242 we have changed the `AccessDeniedException` to the `AuthenticationException`. And the `AccessDeniedException` had logged the errors as info.

akkie:
@see #299

akkie:
Thanks, for submitting this issue

Issue #296, opened 2015-03-02
Allow multiple redirectURL for one provider.


hanfeisun:
In the OAuth2Settings, there is a field called redirectURL, which points to the URL after a successful authentication on the OAuth provider. `

However, sometimes I have two front-end sites (`test-site1` and `test-site2`) using the same play-silhouette back-end application. They will send silhouette different `redirectURL` value in the AJAX call. But currently, silhouette seems to only allow one redirectURL..

I was wondering whether there is a way to allow multiple redirectURL in OAuth2Settings..

akkie:
This is currently not possible. The problem is that the dependencies, based on the chosen DI approach, will be wired either on compile time or on application start. So there is no change to chose the `redirectURL` based on the request. And I've currently no idea how this can be implemented without affecting the other 99,9 percent of the users which haven't such a use case.

Any suggestions?

hanfeisun:
Or is there a way to allow redirectURL to any sites by setting redirectURL to `*`..

I know this may be super unsafe, but it is ok to use it in an internal testing environment..

hanfeisun:
As I understand, silhouette doesn't have to know the `redirectURL` during compile stage or on application start because `redirectURL` is already included as a field in the AJAX request body to `authenticate/:provider`.. Is this true?

akkie:
Yes, in your special case or with the use of Satellizer this may be true. But this isn't the default case.

I think if we implement such a feature then the application could be vulnerable. Please see:
http://security.stackexchange.com/questions/44214/what-is-the-purpose-of-oauth-2-0-redirect-uri-checking
http://homakov.blogspot.de/2013/03/oauth1-oauth2-oauth.html

A quote from the last link under the point Vector 2:
> redirect_uri should be a constant value in application settings.

hanfeisun:
I agree that it will lead to be vulnerable. And this issue may be worked around by setting two different `redirectURL` in different `.conf` file.

but I still think it will be better if developer are able to make authentication easier in development environment by setting it to `*` if they know clearly what they are doing.. Just like `Access-Control-Allow-Origin: *` field in HTTP header.

akkie:
If you set the redirectURL to `*` then the provider would not know where it should redirect to. So this wouldn't work.

I think at this moment the only solution would be to create different environments with different controllers and different configurations. Each for every redirect URL.

In the next version we could allow to override the global config on a per request base. Maybe in 2.1.

hanfeisun:
[...]

PR #304, opened 2015-03-06
Relative callback urls?

igorbernstein:
Please excuse this incomplete PR, its only meant to start a discussion.  I was wondering if you would be open to the idea of allowing relative callback urls in OAuth1, OAuth2 & OpenID settings?  I figure that
- the urls should be configured using play's reverse routes
- the only thing that changes between environments is the domain
- in almost all cases, the domain that starts the oauth flow is the one that will get the redirect from the 3rd party provider
- having less duplication in the conf files is a good thing

Expected usage:
```scala
def provideFacebookProvider(httpLayer: HTTPLayer, stateProvider: OAuth2StateProvider): FacebookProvider = {
    FacebookProvider(httpLayer, stateProvider, OAuth2Settings(
      //...
      redirectURL = routes.SocialAuthController.authenticate("facebook").toString
     //...
  }
```

My current approach should be backwards compatible with absolute urls.

If you are interested in this & are ok with the approach, I'll add tests & doc

Best,
Igor

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/impl/providers/OpenIDProvider.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/providers/OpenIDProvider.scala
@@ -59,7 +59,7 @@ abstract class OpenIDProvider(httpLayer: HTTPLayer, service: OpenIDService, sett
       case None =>
         // Either we get the openID from request or we use the provider ID to retrieve the redirect URL
         val openID = request.extractString(OpenID).getOrElse(settings.providerURL)
-        service.redirectURL(openID).map { url =>
+        service.redirectURL(openID, resolveCallbackURL(settings.callbackURL)).map { url =>
           val redirect = Results.Redirect(fix3749(url))
           logger.debug("[Silhouette][%s] Redirecting to: %s".format(id, url))
           Left(redirect)
@@ -121,9 +121,10 @@ trait OpenIDService {
    * Retrieve the URL where the user should be redirected to start the OpenID authentication process.
    *
    * @param openID The OpenID to use for authentication.
+   * @param resolvedCallbackURL The full callback URL to the application after a successful authentication.
    * @return The redirect URL where the user should be redirected to start the OpenID authentication process.
    */
-  def redirectURL(openID: String): Future[String]
+  def redirectURL(openID: String, resolvedCallbackURL: String): Future[String]
 
   /**
    * From a request corresponding to the callback from the OpenID server, check the identity of the current user.
@@ -140,6 +141,7 @@ trait OpenIDService {
  *
  * @param providerURL The OpenID provider URL used if no openID was given. @see https://willnorris.com/2009/07/openid-directed-identity-identifier-select
  * @param callbackURL The callback URL to the application after a successful authentication on the OpenID provider.
+ *                    The URL can be a relative path which will be resolved against the current request's host.
  * @param axRequired Required attributes to return from the provider after a successful authentication.
  * @param axOptional Optional attributes to return from the provider after a successful authentication.
  * @param realm An URL pattern that represents the part of URL-space for which an OpenID Authentication request is valid.
--- a/silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialProvider.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialProvider.scala
@@ -15,13 +15,15 @@
  */
 package com.mohiva.play.silhouette.impl.providers
 
+import java.net.URI
+
 import com.mohiva.play.silhouette.api.services.AuthInfo
 import com.mohiva.play.silhouette.api.util.ExtractableRequest
 import com.mohiva.play.silhouette.api.{ LoginInfo, Provider }
 import com.mohiva.play.silhouette.impl.exceptions.ProfileRetrievalException
 import com.mohiva.play.silhouette.impl.providers.SocialProfileBuilder._
 import play.api.libs.concurrent.Execution.Implicits._
-import play.api.mvc.Result
+import play.api.mvc.{ RequestHeader, Result }
 
 import scala.concurrent.Future
 
@@ -61,6 +63,22 @@ trait SocialProvider extends Provider with SocialProfileBuilder {
       case e if !e.isInstanceOf[ProfileRetrievalException] =>
         Future.failed(new ProfileRetrievalException(UnspecifiedProfileError.format(id), e))
     }
+  }
+
+  /**
+   * Resolves the url to be absolute relative to the request.
+   *
+   * This will pass the url through if its already absolute.
+   *
+   * @param url The url to resolve.
+   * @param request The current request.
+   * @return The absolute url.
+   */
+  protected def resolveCallbackURL(url: String)(implicit request: RequestHeader): String = URI.create(url) match {
+    case uri if uri.isAbsolute => url
+    case uri =>
+      val scheme = if (request.secure) "https://" else "http://"
+      URI.create(scheme + request.host + request.path).resolve(uri).toString
   }
 }
 
--- a/silhouette/test/com/mohiva/play/silhouette/impl/providers/OAuth1ProviderSpec.scala
+++ b/silhouette/test/com/mohiva/play/silhouette/impl/providers/OAuth1ProviderSpec.scala
@@ -21,8 +21,9 @@ import com.mohiva.play.silhouette.impl.providers.OAuth1Provider._
 import org.specs2.matcher.ThrownExpectations
 import org.specs2.mock.Mockito
 import org.specs2.specification.Scope
-import play.api.mvc.Result
+import play.api.mvc.{ Results, Result }
 import play.api.test.{ FakeRequest, WithApplication }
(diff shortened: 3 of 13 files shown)


akkie:
I think this is a good idea. Maybe it can also solve issue https://github.com/mohiva/play-silhouette/issues/296 in an elegant way by running the same instance on multiple sub domains.

igorbernstein:
@akkie I could use a hint as to how I can add tests for this...I was hoping to add tests to OAuth1ProviderSpec that make sure that:
- provider.authenticate() when configured with a relative url, will include the absolute url(relative to the request host) as a query param
- that provider.authenticate() when configured with an absolute url, will not mangle it

The issue I'm hitting is that the OAuth1Settings & the OAuth1Provider instances are configured in the subclass specs (via the Context) and it doesn't appear that I can modify the OAuth1Settings to try the 2 scenarios. I'll have the same issue with OAuth2 & OpenID providers.
 
Maybe my approach can be improved to make it easier to test? or maybe I'm missing something int the test hierarchy?

akkie:
You can use a spy for the settings. See https://github.com/mohiva/play-silhouette/blob/master/silhouette/test/com/mohiva/play/silhouette/impl/providers/oauth2/VKProviderSpec.scala#L151

igorbernstein:
@akkie This should be ready for review...My only concern with this approach is that it provides no support for resolving a relative url for Clef's data-redirect-url attribute.

silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialProvider.scala
```diff
@@ -62,6 +64,29 @@ trait SocialProvider extends Provider with SocialProfileBuilder {
...
+      val baseUri = if (request.secure) {
+        new URI(s"https://${request.host}")
+      } else {
+        new URI(s"http://${request.host}")
+      }
+
+      baseUri.resolve(url).toString
+    }
+  }
 }
```
akkie: This should be protected and more scalish. Something like this:
```scala
protected def resolveCallbackURL(url: String)(implicit request: RequestHeader) = new URI(url) match {
  case u if u.isAbsolute => u
  case _ => new URI(s"http${if (request.secure) "s"}://${request.host}").resolve(url).toString
}
```

silhouette/test/com/mohiva/play/silhouette/impl/providers/OAuth1ProviderSpec.scala
```diff
@@ -78,6 +79,19 @@ abstract class OAuth1ProviderSpec extends SocialProviderSpec[OAuth1Info] {
...
+      val resolvedCallbackURL = "http://www.example.com/callback-url"
+
+      c.oAuthService.retrieveRequestToken(resolvedCallbackURL) returns Future.successful(c.oAuthInfo)
+      c.oAuthService.redirectUrl(c.oAuthInfo.token) returns c.oAuthSettings.authorizationURL
+      c.oAuthTokenSecretProvider.build(any)(any) returns Future.successful(c.oAuthTokenSecret)
+      c.oAuthTokenSecretProvider.publish(any, any)(any) returns Results.Redirect(c.oAuthSettings.authorizationURL)
+
+      await(c.provider.authenticate())
+      there was one(c.oAuthService).retrieveRequestToken(resolvedCallbackURL)
+    }
```
akkie: You should also test the HTTPS version. The HTTPS test could also be tested without the leading slash to see if a slash will be added to the URL.

silhouette/test/com/mohiva/play/silhouette/impl/providers/OAuth2ProviderSpec.scala
```diff
@@ -109,6 +110,36 @@ abstract class OAuth2ProviderSpec extends SocialProviderSpec[OAuth2Info] {
...
+
+          result(c.provider.authenticate()) {
+            case result =>
+              redirectLocation(result) must beSome.which { url =>
+                url must contain(s"$RedirectURI=${encode(resolvedRedirectURL, "UTF-8")}")
+              }
+          }
+      }
+    }
+
```
akkie: Same as above

silhouette/test/com/mohiva/play/silhouette/impl/providers/OpenIDProviderSpec.scala
```diff
@@ -91,6 +92,18 @@ abstract class OpenIDProviderSpec extends SocialProviderSpec[OpenIDInfo] {
...
+      val callbackURL = "/callback-url"
+      val resolvedCallbackURL = s"http://${req.host}$callbackURL"
+
+      c.openIDSettings.callbackURL returns callbackURL
+      c.openIDService.redirectURL(any, any) returns Future.successful(c.openIDSettings.providerURL)
+
+      await(c.provider.authenticate())
+      there was one(c.openIDService).redirectURL(any, ===(resolvedCallbackURL))
+    }
+
```
akkie: Same as above

akkie:
Seems good to me, expect the parts I've commented. Could you also please [rebase the commits](http://docs.openshift.org/origin-m4/oo_contributors_guide.html#combining-multiple-commits) into a single one. Otherwise good work. I would like to integrate this PR into the next RC2 release.

silhouette/app/com/mohiva/play/silhouette/impl/providers/OpenIDProvider.scala
```diff
@@ -140,6 +141,7 @@ trait OpenIDService {
  *
  * @param providerURL The OpenID provider URL used if no openID was given. @see https://willnorris.com/2009/07/openid-directed-identity-identifier-select
  * @param callbackURL The callback URL to the application after a successful authentication on the OpenID provider.
+ *                    The URL can be a relative path which will be resolved using against a request's host.
```
akkie: I'm not an English pro. But does this really means "be resolved using against" and not "be resolved against" without the "using"?

I think it sounds better:
The URL can be a relative path which will be resolved against the current request's host.

What do you think?

silhouette/test/com/mohiva/play/silhouette/impl/providers/OAuth1ProviderSpec.scala
```diff
@@ -78,6 +79,19 @@ abstract class OAuth1ProviderSpec extends SocialProviderSpec[OAuth1Info] {
           redirectLocation(result) must beSome.which(_ == c.oAuthSettings.authorizationURL)
       }
     }
+    "resolves relative redirectURLs  before starting the flow" in new WithApplication {
```
akkie: Double spaces between `redirectURLs  before`

igorbernstein:
@akkie I think fixed all of the issues you found. If the changes look good to you then I'll squash the PR into a single commit.

igorbernstein:
Sorry, I should've thought of this earlier...
Unless I'm mistaken, 99% of the time the callbackURL/redirectURLs will be the same as the current request's uri. Would it make sense to drop the callbackURL/redirectURL attributes from *Setting classes and instead add a `callbackURL(implicit request: RequestHeader)` method to SocialProvider that by default returns request.uri.  And if a developer encounters a corner case where they need to override the callbackURL, they can subclass the provider & override the method?

What do you think?

igorbernstein:
A couple more thoughts:
The only use case I can think of where current request's url will differ from the callbackURL is if the developer wants to split off the initial call to authenticate from the final call (upon returning from the social provider). Is this something you want to support? I not, the whole thing can be simplified by always using the request's uri. If yes, it might be better to change the SocialProvider api to have the methods initializeAuthentication(callbackUrl:String) & finalizeAuthentication()

igorbernstein:
Nevermind, I was wrong...
We can't depend on the `request.uri` for redirectURL. After the 3rd party redirects back to the `redirectURL`, it will include additional query params, which will make `request.uri` diverge from the original `redirectURL`, which in turn will break `getAccessToken`.

Sorry for all the noise...let me know if PR is good to go & I'll squash

akkie:
Looks very good to me. One thing I would change is the position of your helper methods in the tests. Normally they should be placed in the context instead directly in the specs.

```scala
trait OpenIDProviderSpecContext extends Scope with Mockito with ThrownExpectations {
...
  /**
   * The provider to test.
   */
  def provider: OpenIDProvider

  /**
   * Helper method to ...
   */
  def verifyRelativeCallbackURLResolution(callbackURL: String, secure: Boolean, resolvedCallbackURL: String) = {
     ...
  }
}
```

Otherwise, thumbs up! :thumbsup:

igorbernstein:
I squashed the commits, however I don't think I can move the test helpers into the Context. The test helpers depend on `SocialProviderSpec` for `result`, `FutureAwaits` for `await` and `PlayRunners` for `GET`.  All of those traits are mixed into the spec.  It feels weird remix those traits into the Context. My original intention was to [datatables](https://etorreborre.github.io/specs2/guide/SPECS2-3.0.1/org.specs2.guide.UseDatatables.html), but I ran into issues where the context wasn't being recycled or losing the ability to use `skipped`.
Let me know if I'm missing something or if I misunderstood your comment.

akkie:
Yes, you're right. I thought the context is also located in the specification. Good job!

Issue #312, opened 2015-03-15
EventBus trait?


igorbernstein:
Hi Christian,

I was wondering how you would feel about extracting out an EventBus trait that contained only the `publish(event: Event): Unit ` method, then using that trait in `Environment` and renaming the existing `EventBus` to `impl.util.AkkaEventBus`? The benefits of this change are:

- developers can plug in their own implementations based on other pubsub systems
- testkit can easily implement an empty `FakeEventBus` that doesn't require initializing & shutting down akka for every unit test
- `Environment` can be defined as the minimal set of dependencies needed by controllers with `SecuredAction`s

If this is something you'd be interested in I can put together a PR

akkie:
Sounds good to me!

But what do you mean with "Environment can be defined as the minimal set of dependencies needed by controllers with SecuredActions"? Isn't this exactly so implemented?

igorbernstein:
Ok, great! I should have something to show on Tuesday.

I just meant that all the Apis exposed by environment are very lean & geared for SecuredActions except for EventBus which adds unnecessary subscription functionality.

PR #314, opened 2015-03-16
Extract EventBus trait

igorbernstein:
This PR addresses #312.  It:
- renames `com.mohiva.play.silhouette.api.EventBus` to `com.mohiva.play.silhouette.impl.util.AkkaEventBus`
- creates a new trait `com.mohiva.play.silhouette.api.EventBus`
- adds a noop `com.mohiva.play.silhouette.test.FakeEventBus`
- simplifies SilhouetteSpec to use Mockito mocks & verifications instead of an Akka TestProbe

### DIFF
--- a/silhouette-testkit/app/com/mohiva/play/silhouette/test/Fakes.scala
+++ b/silhouette-testkit/app/com/mohiva/play/silhouette/test/Fakes.scala
@@ -194,6 +194,18 @@ object FakeAuthenticator {
 }
 
 /**
+ * A fake event bus that discards all events.
+ */
+case class FakeEventBus() extends EventBus {
+
+  /**
+   * Publishes the event.
+   * @param event The event to publish
+   */
+  override def publish(event: SilhouetteEvent): Unit = ()
+}
+
+/**
  * A fake environment implementation.
  *
  * @param identities A list of (login info -> identity) pairs to return inside a Silhouette action.
@@ -205,7 +217,7 @@ object FakeAuthenticator {
 case class FakeEnvironment[I <: Identity, T <: Authenticator: TypeTag](
   identities: Seq[(LoginInfo, I)],
   providers: Map[String, Provider] = Map(),
-  eventBus: EventBus = EventBus())
+  eventBus: EventBus = FakeEventBus())
   extends Environment[I, T] {
 
   /**
--- a/silhouette/app/com/mohiva/play/silhouette/api/EventBus.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/api/EventBus.scala
@@ -15,8 +15,6 @@
  */
 package com.mohiva.play.silhouette.api
 
-import akka.event.{ SubchannelClassification, ActorEventBus, LookupClassification }
-import akka.util.Subclassification
 import play.api.i18n.Lang
 import play.api.mvc.RequestHeader
 
@@ -84,51 +82,13 @@ case class NotAuthenticatedEvent(request: RequestHeader, lang: Lang) extends Sil
 case class NotAuthorizedEvent[I <: Identity](identity: I, request: RequestHeader, lang: Lang) extends SilhouetteEvent
 
 /**
- * An event bus implementation which uses a class based lookup classification.
+ * The EventBus api, used to publish authentication related events.
  */
-class EventBus extends ActorEventBus with SubchannelClassification {
-  type Classifier = Class[_ <: SilhouetteEvent]
-  type Event = SilhouetteEvent
+trait EventBus {
 
   /**
-   * The logic to form sub-class hierarchy
+   * Publishes the event.
+   * @param event The event to publish
    */
-  protected implicit val subclassification = new Subclassification[Classifier] {
-    def isEqual(x: Classifier, y: Classifier): Boolean = x == y
-    def isSubclass(x: Classifier, y: Classifier): Boolean = y.isAssignableFrom(x)
-  }
-
-  /**
-   * Publishes the given Event to the given Subscriber.
-   *
-   * @param event The Event to publish.
-   * @param subscriber The Subscriber to which the Event should be published.
-   */
-  protected def publish(event: Event, subscriber: Subscriber): Unit = subscriber ! event
-
-  /**
-   * Returns the Classifier associated with the given Event.
-   *
-   * @param event The event for which the Classifier should be returned.
-   * @return The Classifier for the given Event..
-   */
-  protected def classify(event: Event): Classifier = event.getClass
-}
-
-/**
- * A global event bus instance.
- */
-object EventBus {
-
-  /**
-   * Holds the global event bus instance.
-   */
-  private lazy val instance = new EventBus
-
-  /**
-   * Gets a global event bus instance.
-   *
-   * @return A global event bus instance.
-   */
-  def apply() = instance
+  def publish(event: SilhouetteEvent): Unit
 }
--- a/silhouette/app/com/mohiva/play/silhouette/impl/util/AkkaEventBus.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/util/AkkaEventBus.scala
@@ -1,0 +1,70 @@
+/**
+ * Copyright 2015 Mohiva Organisation (license at mohiva dot com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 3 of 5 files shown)


silhouette-testkit/app/com/mohiva/play/silhouette/test/Fakes.scala
```diff
@@ -194,6 +194,17 @@ object FakeAuthenticator {
 }
 
 /**
+ * A fake event bus that discards all events.
+ */
+case class FakeEventBus() extends EventBus {
+  /**
```
akkie: There should be a newline between the class definition and the comment

silhouette/app/com/mohiva/play/silhouette/api/EventBus.scala
```diff
@@ -84,51 +82,14 @@ case class NotAuthenticatedEvent(request: RequestHeader, lang: Lang) extends Sil
 case class NotAuthorizedEvent[I <: Identity](identity: I, request: RequestHeader, lang: Lang) extends SilhouetteEvent
 
 /**
- * An event bus implementation which uses a class based lookup classification.
+ * The EventBus api, used to publish authentication related events.
  */
-class EventBus extends ActorEventBus with SubchannelClassification {
-  type Classifier = Class[_ <: SilhouetteEvent]
+trait EventBus {
   type Event = SilhouetteEvent
```
akkie: The type isn't really needed here

akkie:
There is a big problem in this implementation because in the application code you must cast the event bus returned from the environment. A `publish` method is definitely not enough. The other problem is that if we additionally define `subscribe` and `unsubscribe`, then in some cases that would still not be enough.  We could add an upper type bound for the `EventBus` to the `Environment`. But then we had three raw types to define on every controller.

igorbernstein:
Hi Christian,

I not entirely sure I agree/understand what you mean.  Unless I'm missing something, the interface for EventBus should be sufficient for authentication.  I understand that it doesn't provide an api for subscription, but that shouldn't be a concern for authentication handling code in Silhouette.  I think EventBus trait is similar to the IdentityService... IdentityService doesn't expose an api for registering or saving users but that's fine because that lies outside the scope of authentication...the developer will add to that api as they see fit when they implement a registration controller.  


EventBus provides just enough api for authentication to work there nothing preventing addition of methods when implementing it.  AkkaEventBus is just an example implementation that provides that functionality + the ability to subscribe to events. Developers can depend directly on that or they implement their own or they can implement a simple adapter that allows for silhouette to inject events into their existing pubsub solution.




Please let me know if I missed your point entirely.




Best,

Igor

igorbernstein:
I think the point I'm trying to make is codified in [Interface segregation principle](Interface segregation principle) of [SOLID](http://en.wikipedia.org/wiki/SOLID_%28object-oriented_design%29)

akkie:
Sure, but then we shouldn't provide a base implementation of the `EventBus` trait, because in this implementation it's useless, because of the fact that you must either cast the event bus to the appropriate implementation or you define the concrete implementation as injection point.

```scala
class Controller @Inject() (eventBus: EventBus) {
  // eventBus has only one method here, so you must cast to AkkaEventBus
}
```

```scala
class Controller @Inject() (eventBus: AkkaEventBus) {
  // eventBus has all methods, but it's useless to define the concrete implementation 
  // here, because it violates against the Liskov substitution principle
}
```

I hope I could clarify my point.

igorbernstein:
I'm sorry to be taking so much of your time, but would you mind explaining how the second example violates LSP? Technically & semantically AkkaEventBus extends but doesn't modify the EventBus behavior.  It can be used anywhere where we need to publish Events.

Is your concern that AkkaEventBus needs an AkkaEventConsumer role trait that exposes methods for subscribing to events?

As always, thank you for your time.

igorbernstein:
One more question:
Do you have an idea as to how I should improve this PR or are you saying that its worth pursuing this refactoring at all?

akkie:
If you define the `AkkaEventBus` as base type, then you cannot easily replace this implementation by another pub sub implementation, because your code depends on the methods of the `AkkaEventBus` implementation.

akkie:
I think the question should be: Is the Akka event bus not the best implementation at all? Or is there a need for another bub sub implementation? I'm not entirely sure about that.

If we decide to implement a base `EventBus` trait, then this trait needs methods like `subscribe` and `unsubscribe`. Or we do not provide an `AkkaEventBus` implementation and let the user define it's own event behavior. In this case he must only implement our base `EventBus` trait, so that his code is compatible with our.

@rfranco Any suggestions?

igorbernstein:
Please correct me if I misunderstood what you are saying:




- LSP was broken by the EventBus class before this PR. Silhouette defines EventBus as a base type and you cannot easily replace this implementation by another pubsub implementation.

- this PR doesn't go far enough to fix it.  This PR fixes the publish side by allowing an easy substitution of the publish method implementation. But the consumer side can't be easily be replaced.




Is that correct?




If so, I'm not entirely sure how you can accommodate easy implementation substitution for the consumer api while maintaining the current api of subscribing via ActorRefs and Classifiers. Any suggestions?




The goal I was trying to achieve was to move the LSP violation slightly further back away from the Silhouette core.  I haven't done any work with akka, but I don't think anyone will need to replace akka implementation if they are using akka to begin with.  The main substitution point I see is to replace akka as a whole, which this PR makes fairly easy

igorbernstein:
Looks like I'm a comment behind.  What about packaging AkkaEventBus as a separate silhouette-akka jar? Or is that overkill?

akkie:
> LSP was broken by the EventBus class before this PR. Silhouette defines EventBus as a base type and you cannot easily replace this implementation by another pubsub implementation.

I'm not sure if LSP was broken before, because LSP presupposes that there is a sub type correlation between to types. You also don't say that Scala is broken by LSP because you cannot easily replace the `String` implementation by another.

But the other is correct.

> Looks like I'm a comment behind. What about packaging AkkaEventBus as a separate silhouette-akka jar? Or is that overkill?

Yea, this is overkill.

rfranco:
I agreed with @akkie. 
My opinion is that Silhouette is a Play's library and Play use native Akka so it should use Akka EventBus. What can be done is create a trait EventBus that extends akka.EventBus then implement some types of eventbus with different Classifier and Subscriber. 

Example
```scala
trait EventBus extends akka.event.EventBus {
  type Event = SilhouetteEvent
  type Classifier = Class[_ <: SilhouetteEvent]
}

class ActorEventBus extends EventBus with  akka.event.ActorEventBus {
....
}
```

And dont make sense move akka implemetation to other project if Play use that.
Maybe `Events` and `EventBus` can be move to a new package `com.mohiva.play.silhouette.api.events`.

@akkie WDYT?

igorbernstein:
> I'm not sure if LSP was broken before, because LSP presupposes that there is a sub type correlation between to types. You also don't say that Scala is broken by LSP because you cannot easily replace the String implementation by another.

That argument can be used for AkkaEventBus as well.  It's existence doesn't incur a  LSP violation, but as you pointed out, substituting its usage in client code will. 




The disadvantages of only client code depending on a concrete implementation of AkkaEventBus (to support subscription) already exists for both Silhouette and client code depending on a concrete implementation of EventBus for both publishing and subscribing.




The overall win for this PR is that developers get an added extension point to plug in a different publishing implementation. And IMHO the negative you pointed out remains the same, just moved to a different class.

igorbernstein:
Sorry for the delayed responses,  I'm traveling today so I have intermittent Internet access.




I don't think abstracting further over akka gives any advantages. The options in order of mymy preference are:

1. This PR

2. Leave everything as is

3. Remove akka impl

4. Further abstractions over akka




I don't think I have any further arguments to support my case, so I'll stay out of further discussion on this issue.




Thank you both for the attention that you have given it.

akkie:
I think the proposal by @rfranco is the better approach. Our default implementation would depend entirely on Akka but the user is open to implement a custom solution based on the interface provided by Akka. We could also create a FakeEventBus as suggested by @igorbernstein

Normally it's the same approach as sugegsted by @igorbernstein, with the exception that we do not provide our own trait, but rather use the trait provided by Akka.

What are you think guys?

akkie:
@igorbernstein Would you agree to make the changes to the pull request based on the suggestions made by @rfranco and me?

igorbernstein:
I still disagree with that approach....I would prefer to keep the current implementation in Silhouette as is.

akkie:
OK, then I close this. Thanks for your contributions on this issues. @igorbernstein @rfranco

PR #315, opened 2015-03-17
Allow custom authenticator service from outside of silhouette package

FabienPennequin:
First of all, I'm really impressive by your play module. It's powerful and highly customizable. Congrats! 

So, I started to moving from securesocial to play-silhouette and I have just one problem… I try to write my own authenticator service but it's seems this isn't allowed. There are 4 methods in AuthenticatorService with access restricted to silhouette package: `touch` `update` `renew` `discard`.

I think this isn't intend and I suggest to change the method visibility to public.

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
@@ -128,7 +128,7 @@ trait AuthenticatorService[T <: Authenticator] {
    * @param authenticator The authenticator to touch.
    * @return The touched authenticator on the left or the untouched authenticator on the right.
    */
-  protected[silhouette] def touch(authenticator: T): Either[T, T]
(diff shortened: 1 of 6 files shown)


akkie:
You are welcome! There is a [reason](https://github.com/mohiva/play-silhouette/blob/master/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala#L97) that the access to this methods is prohibited from outside of Silhouette. Because if you use this methods inside a `SecuredAction` or `UserAwareAction` then, the application doesn't work as expected. 

In Scala you could add your own authenticator implementation into the Silhouette namespace, because you are not bound to the "namespace -> file path" rule.

So I'm, not sure if we should remove the visibility, because it can lead to bugs and misusage.

akkie:
@FabienPennequin Does this work for you?

FabienPennequin:
Thanks for your reply! Yes, it works.

Issue #320, opened 2015-03-26
why embed a Future[Result]?


igorbernstein:
https://github.com/mohiva/play-silhouette/blob/98cf59/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala#L79

This isn't a big deal, I was just curious what was the use case for using a Future[Result] param here. Most of the time the result will be a simple redirect which doesn't require a future computation. In the off chance that the result does need to be computed asynchronously wouldn't the developer just add it as another line in the for comprehension before passing it to the embed?

Best,
Igor

akkie:
Yes, I think for the parameter it can be a Result only. But for the return type it is needed because using the method in a for comprehension with other Futures, it must also return a Future. The idea was also that the user should be able to create an own authenticator service and do future related things in this method. The philosophy of Silhouette is it to cover the most thinkable use cases and to be extendable as possible. And why we are a reactive library, we try to use features everywhere, even if currently there is no need for a future.

Cheers,
Christian

igorbernstein:
Sorry I should've been more clear, I'm specifically asking about the param not the return type. Can you clarify if this:

> we try to use features everywhere, even if currently there is no need for a future

applies to the param or only to the return? If it's only for latter, should I open a pr to drop the future from the param list? Or is it not worth it?

akkie:
I'm not really sure about that. What if a user computes the result inside a Future? Then he must flatMap the Future so that he can pass the Result as argument to the embed method. And then inside the embed method we must create a new Future to wrap the result. I think it is better to hook into the Future a user passes, instead of creating a new one.

igorbernstein:
this won't make a big difference one way or the other. It's either an extra future.successful wrapper or an extra flat map in the client code. And one less flat map in silhouette code if you take my suggestion.




The only point I'm trying to make is that current silhouette code is doing extra work to optimize for a less common use case and makes things slightly less readable for a more common use case.




But as I said, this isn't a huge hurdle...just something I noticed while trying to use the library.

akkie:
> The only point I'm trying to make is that current silhouette code is doing extra work to optimize for a less common use case and makes things slightly less readable for a more common use case.

Please could you outline this a bit more?

igorbernstein:
[...]

PR #322, opened 2015-03-28
Don't wrap results in Futures when passing them to authenticators

igorbernstein:
This PR is to address #320.  It's still a work in progress. I still need to take a pass to make sure all is well.

As per #320, I removed the Future wrapper for embed's Result param, which cascaded to update, renew and discard methods as well. As you can see in the PR, it removed the need for a lot of Future wrapping.  The only place where complexity was added is [Silhouette#invokeBlock](https://github.com/igorbernstein/play-silhouette/blob/immediate-authenticator-result-params/silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala#L387-L389) when invoking the handleNotAuthenticated event handler.


akkie:
All `Results.Status(200)` should be replaced with `Results.Ok`. Otherwise looks good to me.

igorbernstein:
Will do tomorrow

akkie:
Looks good!

igorbernstein:
squashed

akkie:
Thanks, good job!

PR #317, opened 2015-03-30
Provider API enhancements 

akkie:
Hi Christian,

I was working on an example Silhouette app based on your seed and ran across an api awkwardness:

Its a bit weird to be lumping all the providers into a single Map...it requires an upcast in SocialAuthController and requires you to handle configuration errors in the CredentialsAuthController  (missing credentials provider).  I was wondering if you would be open to changing the Environment#providers to Environment#requestProviders and updating the seed to inject the CredentialsProvider & SocialProviders in the CredentialsAuthController & SocialAuthController respectfully?

### DIFF
--- a/project/BuildSettings.scala
+++ b/project/BuildSettings.scala
@@ -32,7 +32,7 @@ object BasicSettings extends AutoPlugin {
 
   override def projectSettings = Seq(
     organization := "com.mohiva",
-    version := "2.0",
+    version := "2.1-SNAPSHOT",
     resolvers ++= Dependencies.resolvers,
     scalaVersion := Dependencies.Versions.scalaVersion,
     crossScalaVersions := Dependencies.Versions.crossScala,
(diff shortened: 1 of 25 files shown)


akkie:
A WIP pull request. I've also adapted the [seed template](https://github.com/mohiva/play-silhouette-seed/pull/46). What do you think?

igorbernstein:
Looks good...the only  suggestion I would recommend is to change the registry get methods to return unwrapped Providers and if the provider doesn't exist to throw a ConfigurationException

akkie:
I think this isn't a good idea. Think of the admin interface we have talked about. Or would you like to handle exceptions in a template? In Scala we have `Option` for this cases, so we must not throw exceptions.

igorbernstein:
I was working under the assumption that the client code will want to render the  clef provider in the list with all other social providers. But as you pointed out in the [discussion of the seed](https://github.com/mohiva/play-silhouette-seed/pull/46#discussion_r27430483) I was wrong.  

You are right, we do need to wrap SocialProviders in Options for rendering Clef separately.  This is good to go from my perspective.

PR #323, opened 2015-04-01
Add method to skip the update of an existing authenticator in a Silhouette action

akkie:
In every Silhouette action, an authenticator gets automatically [updated](https://github.com/mohiva/play-silhouette/blob/master/silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala#L275) on every request. This must be done with the result of the action, because every authenticator can update client side data. The problem is now that it isn't possible to manipulate an authenticator manually inside an action, because it gets automatically overridden with the updated authenticator. To work around this issue I've created the methods [discard](https://github.com/mohiva/play-silhouette/blob/master/silhouette/app/com/mohiva/play/silhouette/api/Authenticator.scala#L59) and [renew](https://github.com/mohiva/play-silhouette/blob/master/silhouette/app/com/mohiva/play/silhouette/api/Authenticator.scala#L75) in the current stable version. I've also [limited the access](https://github.com/mohiva/play-silhouette/blob/master/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala#L97) to the authenticator service to prevent the users for misusage.

Now it seems that this isn't enough, because of [this issue](https://groups.google.com/forum/#!topic/play-silhouette/nNz8lh2O-6Y).

My proposal to fix this issue is to implement an additional `skip` method equally to the `discard` and `renew` methods. I'd also remove the access restriction for the authenticator service to fix #315 because this misusage is also possible with this restriction. With the open access to the service, there is no need more for the `discard` and `renew` methods of the authenticator because the service can now be used directly.

As result we had only the `skip` method, which instructs Silhouette to skip the update of the authenticator, so that an manually updated authenticator can be send to the client.

@igorbernstein @rfranco Any suggestions?

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala
@@ -20,6 +20,7 @@
 package com.mohiva.play.silhouette.api
 
 import com.mohiva.play.silhouette.api.exceptions.{ NotAuthenticatedException, NotAuthorizedException }
+import com.mohiva.play.silhouette.api.services.AuthenticatorResult
 import com.mohiva.play.silhouette.api.util.DefaultEndpointHandler
 import play.api.Play
 import play.api.libs.concurrent.Execution.Implicits._
@@ -266,10 +267,7 @@ trait Silhouette[I <: Identity, A <: Authenticator] extends Controller with Logg
     private def handleInitializedAuthenticator[T](authenticator: A, block: A => Future[HandlerResult[T]])(implicit request: RequestHeader) = {
       val auth = env.authenticatorService.touch(authenticator)
       block(auth.extract).flatMap {
-        case hr @ HandlerResult(pr: Authenticator.Discard, _) =>
-          env.authenticatorService.discard(authenticator, pr).map(pr => hr.copy(pr))
-        case hr @ HandlerResult(pr: Authenticator.Renew, _) =>
-          env.authenticatorService.renew(authenticator, pr).map(pr => hr.copy(pr))
+        case hr @ HandlerResult(pr: AuthenticatorResult, _) => Future.successful(hr)
         case hr @ HandlerResult(pr, _) => auth match {
           // Authenticator was touched so we update the authenticator and maybe the result
           case Left(a) => env.authenticatorService.update(a, pr).map(pr => hr.copy(pr))
@@ -292,10 +290,7 @@ trait Silhouette[I <: Identity, A <: Authenticator] extends Controller with Logg
      */
     private def handleUninitializedAuthenticator[T](authenticator: A, block: A => Future[HandlerResult[T]])(implicit request: RequestHeader) = {
       block(authenticator).flatMap {
-        case hr @ HandlerResult(pr: Authenticator.Discard, _) =>
-          env.authenticatorService.discard(authenticator, pr).map(pr => hr.copy(pr))
-        case hr @ HandlerResult(pr: Authenticator.Renew, _) =>
-          env.authenticatorService.renew(authenticator, pr).map(pr => hr.copy(pr))
+        case hr @ HandlerResult(pr: AuthenticatorResult, _) => Future.successful(hr)
         case hr @ HandlerResult(pr, _) =>
           env.authenticatorService.init(authenticator).flatMap { value =>
             env.authenticatorService.embed(value, pr)
--- a/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
@@ -20,9 +20,60 @@
 package com.mohiva.play.silhouette.api.services
 
 import com.mohiva.play.silhouette.api.{ Authenticator, LoginInfo }
-import play.api.mvc.{ Headers, RequestHeader, Result }
+import play.api.libs.iteratee.Enumerator
+import play.api.mvc._
 
 import scala.concurrent.Future
+
+/**
+ * A marker result which indicates that an operation on an authenticator was processed and
+ * therefore it shouldn't updated automatically.
+ *
+ * Due the fact that the update method gets called on every subsequent request to update the
+ * authenticator related data in the backing store and in the result, it isn't possible to
+ * discard or renew the authenticator simultaneously. This is because the "update" method would
+ * override the result created by the "renew" or "discard" method, because it will be executed
+ * as last in the chain.
+ *
+ * As example:
+ * If we discard the session in a Silhouette action then it will be removed from session. But
+ * at the end the update method will embed the session again, because it gets called with the
+ * result of the action.
+ *
+ * @param result The result to wrap.
+ */
+class AuthenticatorResult(result: Result) extends Result(result.header, result.body, result.connection) {
+
+  /**
+   * Creates a new copy of a `AuthenticatorResult`.
+   *
+   * @param header The response header, which contains status code and HTTP headers.
+   * @param body The response body.
+   * @param connection The connection semantics to use.
+   * @return A copy of a `AuthenticatorResult`.
+   */
+  override def copy(
+    header: ResponseHeader = this.header,
+    body: Enumerator[Array[Byte]] = this.body,
+    connection: HttpConnection.Connection = this.connection) = {
+
+    AuthenticatorResult(super.copy(header, body, connection))
+  }
+}
+
+/**
+ * The companion object.
+ */
+object AuthenticatorResult {
+
+  /**
+   * Instantiates a new authenticator result.
+   *
+   * @param result The result to wrap.
+   * @return An authenticator result.
+   */
+  def apply(result: Result) = new AuthenticatorResult(result)
+}
 
 /**
  * Handles authenticators for the Silhouette module.
@@ -76,7 +127,7 @@ trait AuthenticatorService[T <: Authenticator] {
    * @param request The request header.
    * @return The manipulated result.
    */
-  def embed(value: T#Value, result: Result)(implicit request: RequestHeader): Future[Result]
+  def embed(value: T#Value, result: Result)(implicit request: RequestHeader): Future[AuthenticatorResult]
 
   /**
    * Embeds authenticator specific artifacts into the request.
@@ -94,27 +145,6 @@ trait AuthenticatorService[T <: Authenticator] {
    */
   def embed(value: T#Value, request: RequestHeader): RequestHeader
 
-  ///////////////////////////////////////////////////////////////////////////////////////////////
-  //
-  // Internal API
-  //
-  // Due the fact that the update method gets called on every subsequent request to update the
-  // authenticator related data in the backing store and in the result, it isn't possible to
-  // discard or renew the authenticator simultaneously. This is because the "update" method would
-  // override the result created by the "renew" or "discard" method, because it will be executed
-  // as last in the chain.
-  //
-  // As example:
-  // If we discard the session in a Silhouette action then it will be removed from session. But
-  // at the end the update method will embed the session again, because it gets called with the
-  // result of the action.
-  //
-  // So we restrict the access to this methods and force the developer to use the "renew" and
-  // "discard" methods on the Authenticator instance. These methods wrap the result into a marker
-  // result, for which we can execute the appropriate method on this service.
-  //
-  ///////////////////////////////////////////////////////////////////////////////////////////////
-
   /**
    * Touches an authenticator.
    *
@@ -128,7 +158,7 @@ trait AuthenticatorService[T <: Authenticator] {
    * @param authenticator The authenticator to touch.
    * @return The touched authenticator on the left or the untouched authenticator on the right.
    */
-  protected[silhouette] def touch(authenticator: T): Either[T, T]
+  def touch(authenticator: T): Either[T, T]
 
   /**
    * Updates a touched authenticator.
@@ -142,7 +172,7 @@ trait AuthenticatorService[T <: Authenticator] {
    * @param request The request header.
    * @return The original or a manipulated result.
    */
-  protected[silhouette] def update(authenticator: T, result: Result)(implicit request: RequestHeader): Future[Result]
+  def update(authenticator: T, result: Result)(implicit request: RequestHeader): Future[AuthenticatorResult]
 
   /**
    * Renews the expiration of an authenticator.
@@ -156,7 +186,19 @@ trait AuthenticatorService[T <: Authenticator] {
    * @param request The request header.
    * @return The original or a manipulated result.
    */
-  protected[silhouette] def renew(authenticator: T, result: Result)(implicit request: RequestHeader): Future[Result]
+  def renew(authenticator: T, result: Result)(implicit request: RequestHeader): Future[AuthenticatorResult]
+
+  /**
+   * Renews the expiration of an authenticator without embedding it into the result.
+   *
+   * Based on the implementation, the renew method should revoke the given authenticator first, before
+   * creating a new one. If the authenticator was updated, then the updated artifacts should be returned.
+   *
+   * @param authenticator The authenticator to renew.
+   * @param request The request header.
+   * @return The serialized expression of the authenticator.
+   */
+  def renew(authenticator: T)(implicit request: RequestHeader): Future[T#Value]
 
   /**
    * Manipulates the response and removes authenticator specific artifacts before sending it to the client.
@@ -166,7 +208,7 @@ trait AuthenticatorService[T <: Authenticator] {
    * @param request The request header.
    * @return The manipulated result.
    */
-  protected[silhouette] def discard(authenticator: T, result: Result)(implicit request: RequestHeader): Future[Result]
+  def discard(authenticator: T, result: Result)(implicit request: RequestHeader): Future[AuthenticatorResult]
 }
 
 /**
--- a/silhouette/test/com/mohiva/play/silhouette/api/services/AuthenticatorResultSpec.scala
+++ b/silhouette/test/com/mohiva/play/silhouette/api/services/AuthenticatorResultSpec.scala
@@ -1,0 +1,61 @@
+/**
+ * Copyright 2015 Mohiva Organisation (license at mohiva dot com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.mohiva.play.silhouette.api.services
+
+import play.api.mvc.{ Cookie, Results }
+import play.api.test.{ WithApplication, PlaySpecification }
+
+/**
+ * Test case for the [[com.mohiva.play.silhouette.api.services.AuthenticatorResult]] class.
+ */
+class AuthenticatorResultSpec extends PlaySpecification {
+
+  "The `copy` method" should {
+    "return new a new instance of an authenticator result" in {
+      val result = Results.Ok
+      val authenticatorResult = AuthenticatorResult(result)
+
+      authenticatorResult.copy(result.header, result.body, result.connection) must beAnInstanceOf[AuthenticatorResult]
+    }
+  }
+
+  "The `withSession` method" should {
+    "return new a new instance of an authenticator result" in new WithApplication {
+      val result = Results.Ok
+      val authenticatorResult = AuthenticatorResult(result)
+
+      authenticatorResult.withSession("name" -> "value") must beAnInstanceOf[AuthenticatorResult]
+    }
+  }
+
+  "The `withCookies` method" should {
+    "return new a new instance of an authenticator result" in new WithApplication {
+      val result = Results.Ok
+      val authenticatorResult = AuthenticatorResult(result)
+
+      authenticatorResult.withCookies(Cookie("name", "value")) must beAnInstanceOf[AuthenticatorResult]
+    }
+  }
+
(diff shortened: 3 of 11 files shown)


akkie:
A WIP pull request. I'll add more tests later.

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
```diff
@@ -128,7 +150,7 @@ trait AuthenticatorService[T <: Authenticator] {
    * @param authenticator The authenticator to touch.
    * @return The touched authenticator on the left or the untouched authenticator on the right.
    */
-  protected[silhouette] def touch(authenticator: T): Either[T, T]
+  def touch(authenticator: T): Either[T, T]
```
igorbernstein: Is touch still necessary? can it be absorbed by renew?

silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala
```diff
@@ -266,10 +267,7 @@ trait Silhouette[I <: Identity, A <: Authenticator] extends Controller with Logg
     private def handleInitializedAuthenticator[T](authenticator: A, block: A => Future[HandlerResult[T]])(implicit request: RequestHeader) = {
       val auth = env.authenticatorService.touch(authenticator)
       block(auth.extract).flatMap {
-        case hr @ HandlerResult(pr: Authenticator.Discard, _) =>
-          env.authenticatorService.discard(authenticator, pr).map(pr => hr.copy(pr))
-        case hr @ HandlerResult(pr: Authenticator.Renew, _) =>
-          env.authenticatorService.renew(authenticator, pr).map(pr => hr.copy(pr))
+        case hr @ HandlerResult(pr: AuthenticatorResult, _) => Future.successful(hr)
         case hr @ HandlerResult(pr, _) => auth match {
```
igorbernstein: This is probably a dumb question, but why do you need to double wrap the result?
HandlerResult(AuthenticatorResult(result)) ?

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
```diff
@@ -20,9 +20,52 @@
...
+ */
+object AuthenticatorResult {
+
+  /**
+   * Converts a [[Result]] into [[AuthenticatorResult]]
+   */
+  implicit def toAuthenticatorResult(result: Result): AuthenticatorResult = apply(result)
+  implicit def toAuthenticatorResultF(result: Future[Result]): Future[AuthenticatorResult] = result.map(apply)
+  implicit def toAuthenticatorResultT(result: Try[Result]): Try[AuthenticatorResult] = result.map(apply)
+
```
igorbernstein: Is there a benefit to subclassing Result and providing implicit conversions? I feel like this could lead to some hard to debug issues where a developer accidentally converts a Result to an AuthenticatorResult or loses the marker by doing an authenticatorResult.copy

akkie: I'm not sure if I understand you correct here? If we do not subclass `Result` then we would not return a real `Result` and the user must unwrap the result before he sends it to the client.

With the implicit conversion we let the compiler wrap the results passed to the authenticator methods instead of wrapping it by hand. For the end user there is no difference.

> loses the marker by doing an authenticatorResult.copy

What do you mean here. Please could you provide an example?

silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala
igorbernstein: I think we might be talking past each other again...
Yes, AuthenticatorResult is necessary...it was the main point of my suggestion.  I'm questioning why you still need HandlerResult, Discard & Renew classes...
The code should directly update the Result....there is no reason to delay it using the Update & Discard markers any more

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
igorbernstein: >  we would not return a real Result and the user must unwrap the result before he sends it to the client.

Why you need to return a Result? I think the result should get unwrapped by the ActionBuilder after it notices the wrapper & skips the auto renewal process.  In the normal flow, client code returns a Result whether or it uses the Silhouette ActionBuilders.  In cases where client code needs to modify the Authenticator it should be wrapped in Silhouette ActionBuilders and return the marker wrapper class to signal that it wants to opt out of the normal renewal process.

> loses the marker by doing an authenticatorResult.copy

By sublassing Result you make it very easy for the developer to call withSession or .withHeader on the marker wrapper, which will strip the marker wrapper off which in almost all cases will not be the intended result.

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
igorbernstein: I don't understand why we need a separate method for touch...why can't renew do this in one step? When do you touch authenticators outside of the renew phase?

silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala
akkie: As pointed out, `HandlerResult` has nothing to do with authenticators. This is a different beast and it's needed for the [request handlers](http://silhouette.mohiva.com/v2.0/docs/endpoints#request-handlers). 

But were do you see Discard & Renew classes?

> The code should directly update the Result....there is no reason to delay it using the Update & Discard markers any more

What do you mean here?

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
akkie: Renew and touch are two different things. Renew discards the current authenitcator, then it creates a new one, initializes it and embeds it into the result. Touch updates only the `idleTimeout` so that the authenticator cannot time out. And if the `idleTimeout` was updated then we must update the authenticator client side. Therefore we must call the `update` method on the authenticator.

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
akkie: Sorry, but I still can't follow you. In a `SecuredAction` or `UserAwareAction` a user must return a `Result` regardless of whether he modify the authenticator or not.

```scala
def securedAction = SecuredAction { implicit request =>
  Ok("result") // returns Ok("result")
}

def securedRenewAction = SecuredAction.async { implicit request =>
  request.authenticator.renew(Ok("result")) // returns AuthenticatorResult(Ok("result"))
}
```

Both methods return here a `Result` with the body "result". On both results you can call `withSession` or `withHeader`.

Please could you be a bit more specific and explain what the problem is?

silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala
igorbernstein: I was right the first time...it was a dumb question...I didn't understand Request Handlers...thanks for the link

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
igorbernstein: I'm hoping to discourage a user from doing this:

```scala
request.authenticator.renew(Ok("result")).withHeader(....)
```
This will have unintended result of stripping the marker.  If the AuthenticatoResult wasn't a Result, it would be a compile time error alerting the developer that they are doing something unintended

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
igorbernstein: Again, my apologies, I completely misunderstood what renew does.  May I suggest to rename it to replace?

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
nadirabid: Sorry, I've trying to follow you guys. Could anyone clarify what "marker" refers to?

igorbernstein: Background:

When a user logs in, silhouette embeds an Authenticator in the http response.  The client (ie. the browser) has to repeat that authenticator in subsequent requests.  To limit the lifetime of the authenticator, silhouette keeps an idleTimeout in authenticator. That idleTimeout gets refreshed for every request to keep an active session from timing out. To make this process transparent to developer code, Siljouette's ActionBuilders handle this automatically: they extract the authenticator from the request, update the idleTimeout and embed the updated authenticator in the response.





The problem occurs when a controller wants to embed a different authenticator in the response.  This PR tries to solve that problem by introducing a wrapper type for the result that will signal to Silhouette's ActionBuilders to not overwrite the newly embedded authenticator.  In this case the marker is called AuthenticatorResult.




Hope that helps

nadirabid: Basically the wrapper is being used as a marker to signal skipping the updating process in the case of renew.

Helps immensely. Thanks @igorbernstein.

igorbernstein:
So I finally got some time to sit down play with the code and understood what @studiodev was saying and why you subclassed Result....with type erasure its impossible for a method to accept functions with different signatures.

This PR is a definite improvement over the current situation, but I'm hoping that we can figure out a better way.  I'll keep brainstorming

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
akkie: Sorry for the late response, but I was on travel at Easter.

> I'm hoping to discourage a user from doing this:

I'm not sure if the type of our `AuthenticatorResult` result will be stripped.

```scala
object test {

  case class Result(body: String) {
    def addSession(): Result = this
  }

  class AuthenticatorResult(result: Result) extends Result(result.body)

  def main(args: Array[String]) {

    val result = Result("test")
    val wrappedResult = new AuthenticatorResult(result)

    wrappedResult.addSession() match {
      case r: AuthenticatorResult => println(r.body)
      case _ => throw new Exception("unexpected")
    }
  }
}
```

This prints out `test`.

igorbernstein: Welcome back! hope you had a nice vacation.
That test doesn't reflect the problem...addSession doesn't copy the Result like the real Result does. Is it considered bad form to override a case class' copy method to preserve the subclass?

akkie: Yea, thanks! Had a nice time with my wife and the kids.

> That test doesn't reflect the problem...addSession doesn't copy the Result like the real Result does.

Ahh, now I see the problem.

> Is it considered bad form to override a case class' copy method to preserve the subclass?

I think yes.

akkie: Do you mean something like this:

```scala
object test {

  case class Result(body: String) {
    def addSession(): Result = copy()
  }

  class AuthenticatorResult(result: Result) extends Result(result.body) {

    override def copy(body: String) = new AuthenticatorResult(super.copy(body))
  }

  def main(args: Array[String]) {

    val result = Result("test")
    val wrappedResult = new AuthenticatorResult(result)

    wrappedResult.addSession() match {
      case r: AuthenticatorResult => println(r.body)
      case _ => throw new Exception("unexpected")
    }
  }
}
```

Mhh, it works and I cannot find any posts suggesting that this is bad practice.

igorbernstein: The only other thing I can think of is to use an action wrapper to unwrap the AuthenticatorResult




the type signature for WithAuthenticatorResult.apply would be:


(Unit => Future[AuthenticatorResult]) => Future[Result]




And be used like so:


```scala


*Action.async {

  WithAuthenticatorResult {

     ....

     authenticatorService.discard(...)

  }


```






What do you think?

igorbernstein: Yep something along those lines

akkie: I would prefer the `copy` method over the action wrapper because of the verbosity of the wrapper.

akkie: I've asked about the override of the copy in the Scala mailing list:
https://groups.google.com/forum/#!topic/scala-user/JUK9PX3ExAc

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
```diff
@@ -20,9 +20,52 @@
...
+/**
+ * The companion object.
+ */
+object AuthenticatorResult {
+
+  /**
+   * Converts a [[Result]] into [[AuthenticatorResult]]
+   */
+  implicit def toAuthenticatorResult(result: Result): AuthenticatorResult = apply(result)
+  implicit def toAuthenticatorResultF(result: Future[Result]): Future[AuthenticatorResult] = result.map(apply)
```
rfranco: My suggested is avoid the implicit import for `ExecutionContext`

```scala
implicit def toAuthenticatorResultF(result: Future[Result])(implicit ec:ExecutionContext): Future[AuthenticatorResult] = result.map(apply)
```

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
rfranco: Is necessary to return a `Either`? i think could be use a `Option`, if is touch it returns `Some` if not `None`.
That avoid misconception

akkie: I think `Either` is better, because it provides a more fluent interface. With `Option` you must store the given authenticator in a variable to access it in case the authenticator was not touched.

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
akkie: This is something on my to-do list for the complete lib.

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
rfranco: I dont get the 'more fluent interface' because there is a `implicit class` to convert Either[T,T] to T

```scala
private implicit class ExtractEither[T](r: Either[T, T]) {
	def extract: T = r.fold(identity, identity)
}
private def handleInitializedAuthenticator[T](authenticator: A, block: A => Future[HandlerResult[T]])(implicit request: RequestHeader) = {
	val auth: Either[T, T] = env.authenticatorService.touch(authenticator)
    block(auth.extract).flatMap {
    	...
    }
}
```

but if it uses a option as return that can be done without implicit convert.

```scala
private def handleInitializedAuthenticator[T](authenticator: A, block: A => Future[HandlerResult[T]])(implicit request: RequestHeader) = {
	val auth: Option[T] = env.authenticatorService.touch(authenticator)
    block(auth.getOrElse(authenticator)).flatMap {
    	...
    }
}
```
I always prefer a code without implicit that's more clean to understand

akkie: You must look through the complete code of the Silhouette class. The `handleAuthentication` method returns also an `Either[Authenticator, Authenticator]`for the old or maybe a new authenticator. So the `extract` method is used five times in this class. Sure we could also use `fold(identity, identity)` directly, but I thought the extract method is a bit shorter.

Remember also that the `touch` method pertains to the public API of the authenticator service. Just because you don't see my statement 'more fluent interface' valid in the scope of the Silhouette class doesn't mean it isn't valid outside of this class.

I don't really see your problem here. If it returns `Option[A]` or `Either[A, A]`, for me it is nearly the same. I see benefits by returning either the unmodified or the modified, so I can work always with the result of the function and not in case of the unmodified with another variable. For me it's more the functional way.

rfranco: I had and have a lot of problem on my team about the understanding of left and right on Either primarily when use `Either` to return the same type because you need to read the code or the documentation to understand what is the left and right.
It's was only one suggestion, if you are ok with Either so keep going.

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
akkie: Mhhh, no answer to my question in the mailing list. Can we agree that this pull request is a solution if we override the `copy` method in the `AuthenticatorResult` class?

igorbernstein: So looking around the internet I don't think overriding the copy method is an issue.  I think the main  issue is that subclassing a case class is considered bad form. From what I've been able to gather, case classes are meant to be leafs of a class hierarchy so that they can hold the equivalence relation. By introducing a subclass of Result we slightly taint that property...equality is no longer straight forward.

http://stackoverflow.com/a/11159036

Having said that, I think its an acceptable price to pay for this

igorbernstein: oops missed your post

> Mhhh, no answer to my question in the mailing list. Can we agree that this pull request is a solution if we override the copy method in the AuthenticatorResult class?

Yep, I say go for it

akkie:
From my perspective this pull request is ready to merge. Any other suggestions?

igorbernstein:
I can look it over tomorrow morning

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
```diff
@@ -20,9 +20,66 @@
...
+
+  /**
+   * Creates a new copy of a `AuthenticatorResult`.
+   *
+   * @param header The response header, which contains status code and HTTP headers.
+   * @param body The response body.
+   * @param connection The connection semantics to use.
+   * @return A copy of a `AuthenticatorResult`.
+   */
+  override def copy(header: ResponseHeader, body: Enumerator[Array[Byte]], connection: HttpConnection.Connection) = {
```
igorbernstein: I think this should have default params:
override def copy(header: ResponseHeader = this.header, body: Enumerator[Array[Byte]]=this.body, connection: HttpConnection.Connection=this.connection)

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
igorbernstein: I still think its dangerous to have implicit conversions here...

1. it can make debugging really difficult
2. makes the code harder to understand because the subclasses don't list their return type, so the reader must look at the trait's signature to figure out whats going on.  This in turn makes implementing your own authenticator less straightforward

The only benefit I see here is saving a few keystrokes

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
akkie: Yep, you are right!

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
akkie: OK, I'll remove the implicit conversions

akkie:
@igorbernstein I've updated the pull request with your suggestions.

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
```diff
@@ -156,7 +186,19 @@ trait AuthenticatorService[T <: Authenticator] {
...
+   * Renews the expiration of an authenticator without embedding it into the result.
+   *
+   * Based on the implementation, the renew method should revoke the given authenticator first, before
+   * creating a new one. If the authenticator was updated, then the updated artifacts should be returned.
+   *
+   * @param authenticator The authenticator to renew.
+   * @param request The request header.
+   * @return The serialized expression of the authenticator.
+   */
+  def renew(authenticator: T)(implicit request: RequestHeader): Future[T#Value]
```
igorbernstein: Sorry, this is probably a dumb question, but I didn't notice the addition of this method earlier. I don't quite understand its purpose and to be honest not sure how it relates to this PR.

silhouette/test/com/mohiva/play/silhouette/api/services/AuthenticatorResultSpec.scala
```diff
@@ -0,0 +1,34 @@
...
+ */
+class AuthenticatorResultSpec extends PlaySpecification {
+
+  "The `copy` method" should {
+    "return new a new instance of an authenticator result" in {
+      val result = Results.Ok
+      val authenticatorResult = AuthenticatorResult(result)
+
+      authenticatorResult.copy(result.header, result.body, result.connection) must beAnInstanceOf[AuthenticatorResult]
+    }
```
igorbernstein: maybe add a spec or 2 to make sure that withSession, withCookie & withHeader return AuthenticatorResult instances?

igorbernstein:
I think this looks great except for the 2 nitpicks that I mentioned (renew overload & AuthenticatorResult specs)...I don't feel strongly about either so feel free to ignore em & merge

silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
akkie: This is because of #325

silhouette/test/com/mohiva/play/silhouette/api/services/AuthenticatorResultSpec.scala
akkie: Good idea!

PR #327, opened 2015-04-08
Authenticator result alternative

### LINKED ISSUE #323: Add method to skip the update of an existing authenticator in a Silhouette action
In every Silhouette action, an authenticator gets automatically [updated](https://github.com/mohiva/play-silhouette/blob/master/silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala#L275) on every request. This must be done with the result of the action, because every authenticator can update client side data. The problem is now that it isn't possible to manipulate an authenticator manually inside an action, because it gets automatically overridden with the updated authenticator. To work around this issue I've created the methods [discard](https://github.com/mohiva/play-silhouette/blob/master/silhouette/app/com/mohiva/play/silhouette/api/Authenticator.scala#L59) and [renew](https://github.com/mohiva/play-silhouette/blob/master/silhouette/app/com/mohiva/play/silhouette/api/Authenticator.scala#L75) in the current stable version. I've also [limited the access](https://github.com/mohiva/play-silhouette/blob/master/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala#L97) to the authenticator service to prevent the users for misusage.

Now it seems that this isn't enough, because of [this issue](https://groups.google.com/forum/#!topic/play-silhouette/nNz8lh2O-6Y).

My proposal to fix this issue is to implement an additional `skip` method equally to the `discard` and `renew` methods. I'd also remove the access restriction for the authenticator service to fix #315 because this misusage is also possible with this restriction. With the open access to the service, there is no need more for the `discard` and `renew` methods of the authenticator because the service can now be used directly.

As result we had only the `skip` method, which instructs Silhouette to skip the update of the authenticator, so that an manually updated authenticator can be send to the client.

@igorbernstein @rfranco Any suggestions?

akkie:
This is an alternative pull request to fix #323. Instead of using a marker `Result` it uses a mutable flag in the `Authenticator` instance. This doesn't work with the embed method, because this method gets no authenticator passed. But it can be used with all other methods which manipulates/creates an authenticator. The other pull request with the marker `Result` works only with the methods that returns a `Result`.

There are some points which can be changed:
- We could allow to set the flag from outside of Silhouette
- We could set the flag also in the `create` and `init` methods

@igorbernstein  @rfranco What do you think?

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/api/Authenticator.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/api/Authenticator.scala
@@ -42,6 +42,23 @@ trait Authenticator {
    * @return True if the authenticator isn't expired and isn't timed out.
    */
   def isValid: Boolean
+
+  /**
+   * A flag which indicates that an operation on an authenticator was processed and
+   * therefore not updated automatically.
+   *
+   * Due the fact that the update method gets called on every subsequent request to update the
+   * authenticator related data in the backing store and in the result, it isn't possible to
+   * discard or renew the authenticator simultaneously. This is because the "update" method would
+   * override the result created by the "renew" or "discard" method, because it will be executed
+   * as last in the chain.
+   *
+   * As example:
+   * If we discard the session in a Silhouette action then it will be removed from session. But
+   * at the end the update method will embed the session again, because it gets called with the
+   * result of the action.
+   */
+  private[silhouette] var skipUpdate = false
 }
 
 /**
--- a/silhouette/app/com/mohiva/play/silhouette/api/Authenticator.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/api/Authenticator.scala
@@ -19,13 +19,6 @@
  */
 package com.mohiva.play.silhouette.api
 
-import com.mohiva.play.silhouette.api.Authenticator.Discard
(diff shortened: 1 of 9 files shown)


igorbernstein:
Just to make sure we are on the same page, I just wanna recap that the underlying problem that you/we are trying to solve here is to "update & embed the request authenticator into the result unless the developer tells you otherwise". 

Please correct me if I'm wrong, but the general idea in this PR is to have the AuthenticatorService methods taint the authenticator in the request via side effects so that the ActionBuilders skip the touch?

If my understanding is correct, then here are the negatives that I see with this approach:

- Its unclear w/o reading the code/documentation that those methods expect the request authenticator.  If someone passes in a copy then all is lost.
- Mutating the request to affect the result feels really wrong. The basic http flow is a function Request => Result. This approach seems like its treating the Request as part of the Result.
- Adding mutable state seems to go against best practices of scala & functional programming.

None of those are awful problems, but what benefits do you gain from this approach vs #323?

akkie:
Your negative points are all valid. This was a try for a solution as described in a [comment](https://github.com/mohiva/play-silhouette/pull/323#issuecomment-88439812) to #323. And I'm with you that the other pull request is a better solution.

Issue #326, opened 2015-04-07
Split AuthInfoService.save interface method into create and update 


nadirabid:
Originally identified in this [post](https://groups.google.com/forum/#!topic/play-silhouette/XvRAfMG4QBM).

Currently, the interface method [`AuthInfoService.save`](https://github.com/mohiva/play-silhouette/blob/master/silhouette/app/com/mohiva/play/silhouette/api/services/AuthInfoService.scala#L46) is used to both create and update an instance of `AuthInfo`. 
 
I'd like to request splitting `AuthInfoService.save` into `AuthInfoService.create` and `AuthInfoService.update`. The main benefits being: 
   * avoid some difficulties in implementing a method that either has to create or update circumstantially 
   * make the interface more clear as it conforms to common CRUD operations

akkie:
We should generally find a consistent naming scheme for the data access layers. Currently we have three types of layers.

* Services -> (AuthInfoService, IdentityService)
* DAOs -> (AuthInfoDAO, AuthenticatorDAO)
* Layers -> CacheLayer

The usage is as follows:
* Layers can be used in DAOs or services to provide access to a single storage technology.
* DAOs can be used in services to persist the data independently from a persistence layer.
* Services are used in the Silhouette API directly to delegate operations to an underlying storage API.

Common terms for data access operations are the following:
* save, store, upsert -> Is used mostly for insert and update operations, consolidated in a single operation
* insert, create -> To insert or create data
* update -> To update data
* find, retrieve, select, read -> To find, retrieve, select or read data
* remove, delete -> To remove or delete data

For me the term `create` implies that if I call such method, that it creates something new and returns this created data. So I personally don't like the term for a method that gets already created data passed to save it in a persistence layer. For the other methods I would suggest to use more common method names, they do not imply a special type of database type. What I mean is that as example the term `select` is for me to much associated with SQL.

So I would suggest the following naming scheme:
* save -> for insert and update in a single operation
* insert
* find/retrieve
* update
* remove

DAOs should not have a `save` method because not every storage system has a consistent method to handle insert and update operations in a single operation.

For DAOs I like more the term `find` and for services I like more the term `retrieve`.

Any suggestions?

nadirabid:
I think follow you thinking in terms of using `create` at the persistence layer. Personally, an API with `insert` and `update` makes the usage clear, so I don't have any objections. 

[...]

PR #334, opened 2015-04-29
Improve the storage API

akkie:
See: https://github.com/mohiva/play-silhouette/issues/326


nadirabid:
Perfect, I'm definitely looking forward to this. Looks like you also decided in favor of the repository pattern, which does the job of keeping the API clear.  

Based on the changes, am I correct in assuming that this change will require Play 2.4?

akkie:
@nadirabid Yes, all new features will go into the next 3.0 release which addresses mainly Play 2.4.

PR #340, opened 2015-05-09
Allow to return a Future in the Authorization.isAuthorized method

GonzaGonza:
isAuthorized method use Futures, so now it is possible to avoid Await when you use a DB (like MongoDb) to know if a user is authorized.


akkie:
Really nice! But please could you rewrite the commit message into something like this: `Allow to return a Future in the Authorization.isAuthorized method`

GonzaGonza:
Sure i will rewrite it, and re-do the pullrequest. Thanks!

akkie:
Thanks!

Issue #341, opened 2015-05-10
Silhouette is not ExecutionContext transparent


joaoraf:
Since there are several places where a Future is created and some callback methods that a user os Silhouette has to implement return a Future, an ExecutionContext is needed at those points.
Since it isn't been propagated from the library calls up to the user, the callback implementations must rely on a ExecutionContext which is not necessarily the same that the library uses, creating a possibly unnecessary thread switch to execute the callback.
I propose that the library should propagate an implicit ExecutionContext through its calls, allowing the callback methods to use if needed.

akkie:
@see #342

PR #342, opened 2015-05-10
Alter the library methods to propagate an implicit ExecutionContext

joaoraf:
Details on issue #341.

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala
@@ -24,10 +24,10 @@ import com.mohiva.play.silhouette.api.services.AuthenticatorResult
 import com.mohiva.play.silhouette.api.util.DefaultEndpointHandler
 import play.api.Play
 import play.api.i18n.{ MessagesApi, I18nSupport }
-import play.api.libs.concurrent.Execution.Implicits._
 import play.api.mvc._
 
 import scala.concurrent.Future
+import scala.concurrent.ExecutionContext
 import scala.language.higherKinds
 
 /**
@@ -76,7 +76,7 @@ trait Silhouette[I <: Identity, A <: Authenticator] extends Controller with Logg
    * @param request The request header.
    * @return The result to send to the client.
    */
-  protected def onNotAuthenticated(request: RequestHeader): Option[Future[Result]] = None
+  protected def onNotAuthenticated(request: RequestHeader)(implicit ec: ExecutionContext): Option[Future[Result]] = None
 
   /**
    * Implement this to return a result when the user is authenticated but not authorized.
@@ -86,7 +86,7 @@ trait Silhouette[I <: Identity, A <: Authenticator] extends Controller with Logg
    * @param request The request header.
    * @return The result to send to the client.
    */
-  protected def onNotAuthorized(request: RequestHeader): Option[Future[Result]] = None
+  protected def onNotAuthorized(request: RequestHeader)(implicit ec: ExecutionContext): Option[Future[Result]] = None
 
   /**
    * Default exception handler for silhouette exceptions which translates an exception into
@@ -98,7 +98,7 @@ trait Silhouette[I <: Identity, A <: Authenticator] extends Controller with Logg
    * @param request The request header.
    * @return The result to send to the client based on the exception.
    */
-  protected def exceptionHandler(implicit request: RequestHeader): PartialFunction[Throwable, Future[Result]] = {
+  protected def exceptionHandler(implicit request: RequestHeader, ec: ExecutionContext): PartialFunction[Throwable, Future[Result]] = {
     case e: NotAuthenticatedException =>
       logger.info(e.getMessage, e)
       handleNotAuthenticated
@@ -119,7 +119,7 @@ trait Silhouette[I <: Identity, A <: Authenticator] extends Controller with Logg
    * @param request The request header.
    * @return The result to send to the client if the user isn't authenticated.
    */
-  private def handleNotAuthenticated(implicit request: RequestHeader): Future[Result] = {
+  private def handleNotAuthenticated(implicit request: RequestHeader, ec: ExecutionContext): Future[Result] = {
     logger.debug("[Silhouette] Unauthenticated user trying to access '%s'".format(request.uri))
 
     onNotAuthenticated(request).orElse {
@@ -143,7 +143,7 @@ trait Silhouette[I <: Identity, A <: Authenticator] extends Controller with Logg
    * @param request The request header.
    * @return The result to send to the client if the user isn't authorized.
    */
-  private def handleNotAuthorized(implicit request: RequestHeader): Future[Result] = {
+  private def handleNotAuthorized(implicit request: RequestHeader, ec: ExecutionContext): Future[Result] = {
     logger.debug("[Silhouette] Unauthorized user trying to access '%s'".format(request.uri))
 
     onNotAuthorized(request).orElse {
@@ -180,7 +180,7 @@ trait Silhouette[I <: Identity, A <: Authenticator] extends Controller with Logg
      * @tparam T The type of the data included in the handler result.
      * @return A handler result.
      */
-    final def apply[T](block: R[AnyContent] => Future[HandlerResult[T]])(implicit request: Request[AnyContent]): Future[HandlerResult[T]] = {
+    final def apply[T](block: R[AnyContent] => Future[HandlerResult[T]])(implicit request: Request[AnyContent], ec: ExecutionContext): Future[HandlerResult[T]] = {
       invokeBlock(block)
     }
 
@@ -193,8 +193,8 @@ trait Silhouette[I <: Identity, A <: Authenticator] extends Controller with Logg
      * @tparam T The type of the data included in the handler result.
      * @return A handler result.
      */
-    final def apply[B, T](request: Request[B])(block: R[B] => Future[HandlerResult[T]]): Future[HandlerResult[T]] = {
-      invokeBlock(block)(request)
+    final def apply[B, T](request: Request[B])(block: R[B] => Future[HandlerResult[T]])(implicit ec: ExecutionContext): Future[HandlerResult[T]] = {
+      invokeBlock(block)(request, ec)
     }
 
     /**
@@ -208,7 +208,7 @@ trait Silhouette[I <: Identity, A <: Authenticator] extends Controller with Logg
      * @tparam T The type of the data included in the handler result.
      * @return A handler result.
      */
-    protected def invokeBlock[B, T](block: R[B] => Future[HandlerResult[T]])(implicit request: Request[B]): Future[HandlerResult[T]]
+    protected def invokeBlock[B, T](block: R[B] => Future[HandlerResult[T]])(implicit request: Request[B], ec: ExecutionContext): Future[HandlerResult[T]]
 
     /**
      * Handles a block for an authenticator.
@@ -221,7 +221,7 @@ trait Silhouette[I <: Identity, A <: Authenticator] extends Controller with Logg
      * @param request The current request header.
      * @return A handler result.
      */
-    protected def handleBlock[T](authenticator: Either[A, A], block: A => Future[HandlerResult[T]])(implicit request: RequestHeader) = {
+    protected def handleBlock[T](authenticator: Either[A, A], block: A => Future[HandlerResult[T]])(implicit request: RequestHeader, ec: ExecutionContext) = {
       authenticator match {
         case Left(a) => handleInitializedAuthenticator(a, block)
         case Right(a) => handleUninitializedAuthenticator(a, block)
@@ -241,10 +241,10 @@ trait Silhouette[I <: Identity, A <: Authenticator] extends Controller with Logg
      * @return A tuple which consists of (maybe the existing authenticator on the left or a
      *         new authenticator on the right -> maybe the identity).
      */
-    protected def handleAuthentication[B](implicit request: Request[B]): Future[(Option[Either[A, A]], Option[I])] = {
+    protected def handleAuthentication[B](implicit request: Request[B], ec: ExecutionContext): Future[(Option[Either[A, A]], Option[I])] = {
       env.authenticatorService.retrieve.flatMap {
         // A valid authenticator was found so we retrieve also the identity
-        case Some(a) if a.isValid => env.identityService.retrieve(a.loginInfo).map(i => Some(Left(a)) -> i)
+        case Some(a) if a.isValid => env.identityService.retrieve(a.loginInfo)(ec).map(i => Some(Left(a)) -> i)
         // An invalid authenticator was found so we needn't retrieve the identity
         case Some(a) if !a.isValid => Future.successful(Some(Left(a)) -> None)
         // No authenticator was found so we try to authenticate with a request provider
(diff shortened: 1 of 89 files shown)


joaoraf:
Forgot to add that it pass all tests. I had to change the tests due to the extra ExecutionContext parameters.

silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala
```diff
@@ -24,10 +24,11 @@ import com.mohiva.play.silhouette.api.services.AuthenticatorResult
 import com.mohiva.play.silhouette.api.util.DefaultEndpointHandler
 import play.api.Play
 import play.api.i18n.{ MessagesApi, I18nSupport }
-import play.api.libs.concurrent.Execution.Implicits._
+////import play.api.libs.concurrent.Execution.Implicits._
```
akkie: This comment can be removed

silhouette/app/com/mohiva/play/silhouette/impl/providers/BasicAuthProvider.scala
```diff
@@ -22,10 +22,11 @@ import com.mohiva.play.silhouette.api.{ LoginInfo, RequestProvider }
 import com.mohiva.play.silhouette.impl.exceptions.{ IdentityNotFoundException, InvalidPasswordException }
 import com.mohiva.play.silhouette.impl.providers.BasicAuthProvider._
 import play.api.http.HeaderNames
-import play.api.libs.concurrent.Execution.Implicits._
+////import play.api.libs.concurrent.Execution.Implicits._
```
akkie: This comment can be removed

silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala
```diff
@@ -27,13 +27,14 @@ import com.mohiva.play.silhouette.api.util.{ ExtractableRequest, HTTPLayer }
 import com.mohiva.play.silhouette.impl.exceptions.{ AccessDeniedException, UnexpectedResponseException }
 import com.mohiva.play.silhouette.impl.providers.OAuth2Provider._
 import com.mohiva.play.silhouette.{ impl, _ }
-import play.api.libs.concurrent.Execution.Implicits._
+////import play.api.libs.concurrent.Execution.Implicits._
```
akkie: This comment can be removed

silhouette/app/com/mohiva/play/silhouette/impl/providers/OpenIDProvider.scala
```diff
@@ -21,10 +21,11 @@ import com.mohiva.play.silhouette.api._
 import com.mohiva.play.silhouette.api.util.{ ExtractableRequest, HTTPLayer }
 import com.mohiva.play.silhouette.impl.exceptions.UnexpectedResponseException
 import com.mohiva.play.silhouette.impl.providers.OpenIDProvider._
-import play.api.libs.concurrent.Execution.Implicits._
+////import play.api.libs.concurrent.Execution.Implicits._
```
akkie: This comment can be removed

silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth1/LinkedInProvider.scala
```diff
@@ -24,10 +24,11 @@ import com.mohiva.play.silhouette.api.util.HTTPLayer
 import com.mohiva.play.silhouette.impl.exceptions.ProfileRetrievalException
 import com.mohiva.play.silhouette.impl.providers._
 import com.mohiva.play.silhouette.impl.providers.oauth1.LinkedInProvider._
-import play.api.libs.concurrent.Execution.Implicits._
+////import play.api.libs.concurrent.Execution.Implicits._
```
akkie: This comment can be removed

silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth1/TwitterProvider.scala
```diff
@@ -24,10 +24,11 @@ import com.mohiva.play.silhouette.api.util.HTTPLayer
 import com.mohiva.play.silhouette.impl.exceptions.ProfileRetrievalException
 import com.mohiva.play.silhouette.impl.providers._
 import com.mohiva.play.silhouette.impl.providers.oauth1.TwitterProvider._
-import play.api.libs.concurrent.Execution.Implicits._
+////import play.api.libs.concurrent.Execution.Implicits._
```
akkie: This comment can be removed

silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth1/XingProvider.scala
```diff
@@ -24,10 +24,11 @@ import com.mohiva.play.silhouette.api.util.HTTPLayer
 import com.mohiva.play.silhouette.impl.exceptions.ProfileRetrievalException
 import com.mohiva.play.silhouette.impl.providers._
 import com.mohiva.play.silhouette.impl.providers.oauth1.XingProvider._
-import play.api.libs.concurrent.Execution.Implicits._
+////import play.api.libs.concurrent.Execution.Implicits._
```
akkie: This comment can be removed

silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth1/services/PlayOAuth1Service.scala
```diff
@@ -22,11 +22,12 @@ package com.mohiva.play.silhouette.impl.providers.oauth1.services
 import com.mohiva.play.silhouette.api.Logger
 import com.mohiva.play.silhouette.impl.providers.oauth1.services.PlayOAuth1Service._
 import com.mohiva.play.silhouette.impl.providers.{ OAuth1Info, OAuth1Service, OAuth1Settings }
-import play.api.libs.concurrent.Execution.Implicits._
+////import play.api.libs.concurrent.Execution.Implicits._
```
akkie: This comment can be removed

silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth2/ClefProvider.scala
```diff
@@ -20,10 +20,11 @@ import com.mohiva.play.silhouette.api.util.HTTPLayer
 import com.mohiva.play.silhouette.impl.exceptions.ProfileRetrievalException
 import com.mohiva.play.silhouette.impl.providers._
 import com.mohiva.play.silhouette.impl.providers.oauth2.ClefProvider._
-import play.api.libs.concurrent.Execution.Implicits._
+////import play.api.libs.concurrent.Execution.Implicits._
```
akkie: This comment can be removed

silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth2/DropboxProvider.scala
```diff
@@ -21,10 +21,11 @@ import com.mohiva.play.silhouette.impl.exceptions.ProfileRetrievalException
 import com.mohiva.play.silhouette.impl.providers._
 import com.mohiva.play.silhouette.impl.providers.oauth2.DropboxProvider._
 import play.api.http.HeaderNames._
-import play.api.libs.concurrent.Execution.Implicits._
+////import play.api.libs.concurrent.Execution.Implicits._
```
akkie: This comment can be removed

akkie:
You are my hero! Really! This was the next task on my to do list and now you save me a lot of time.

There are two things I've noticed. The first are the comments with the old implicit execution context. They are widespread over the complete project. These should be removed. The other is a question regarding the equality matcher you've used in the tests:

```scala
env.identityService.retrieve(===(identity.loginInfo))(any) returns Future.successful(None)
```

Why is it necessary now? Could you also explain what the smart mock does? I'm not aware of such a functionality.

@rfranco Please could you also review the pull request?

joaoraf:
Sorry about the comments. I thought I had removed then before commiting.
I had lots of trouble with the matchers. I actually had to learn about Mockito, which was a good thing.
Using your example, the original expression was:
    env.identityService.retrieve(identity.loginInfo) returns Future.successful(None)
Since there wasn't any matchers inside the expression "env.identityService.retrieve(identity.loginInfo)"
However, now we have a implicit ExecutionContext parameters and we need to add a wildcard (any)
matcher to the expression, otherwise Mockito freaks out:
   env.identityService.retrieve(identity.loginInfo)(any) returns Future.successful(None)
Now mockito complains that there should be two matchers instead of just one. The rule is, if a matcher is
used, then all parameters must be matchers also. Thus we need:
   env.identityService.retrieve(===(identity.loginInfo))(any) returns Future.successful(None)
I hope that explains it.
I will try to amend the commit (I am also been (willingly) forced to learn to use GitHub and git)
I am really happy to be useful. Your library is very useful and the code is extremely elegant.

joaoraf:
About the smart mock: I had some problems with null values being returned from unexpected unstubbed methods after I started adding the ExecutionContext parameters. 
I eventually found out that I had to pass the implicit parameters explicitly in those methods, otherwise Mockito would not know how to stub then and would just return null.
To find these out I used the smart option, which makes the mock return a SmartNull instead of a null value, resulting in a error message which explains which method was unstubbed (see https://code.google.com/p/specs/wiki/UsingMockito#Smart_mocks).

akkie:
Ahhh, now I see. Thanks for the explanation.

rfranco:
@akkie I'm traveling now and i'll be back Tuesday then i can do better review so i did a quick review and looks great anyway maybe it could be done as we did with `MessagesApi`, add the `ExecutionContext` on the `Environment` that dont need to spray the ExecutionContext for everywhere.

akkie:
@rfranco But then we must spread the environment around.
@joaoraf Please could you merge the changes from #340 into this pull request?

joaoraf:
In an hour.

joaoraf:
@akkie @rfranco  The build failed in a test. I don't see what the problem is. In my local code the tests run ok. 

On CookieSecretSpec:

[info] The `publish` method of the provider should
[error] x add the secret to the cookie
[error] 'Some(Cookie(OAuth1TokenSecret,2-YEQIlweDazjnTbHwd3q1M3JhdVAxUcb4EC8IjXQSiZ7XMbSj/I/Bu4bFLFjAAtqORRvgycnEg0AjCQlJ2wglhw==,Some(299),/,None,true,true))' is Some but 'Some(299)' is Some but '299' is not equal to '300' (CookieSecretSpec.scala:125)

joaoraf:
@akkie @rfranco seems to be ready now

akkie:
@joaoraf The test error has to do with https://github.com/playframework/playframework/issues/3120. Maybe we should change the test. Anyway, this shouldn't be a part of this pull request.

Please could you squash the commits into a single commit. Otherwise, for me it looks ready to merge.

akkie:
@joaoraf I think the [Authorization.isAuthorized](https://github.com/mohiva/play-silhouette/blob/master/silhouette/app/com/mohiva/play/silhouette/api/Authorization.scala#L43) method needs also an execution context.

joaoraf:
@akkie everything tested and squashed.

akkie:
@joaoraf The import comments are still available.

joaoraf:
Damned squashing!

rfranco:
I reviewed the code and i'm not sure about add ExecutionContext in everyplace.

One example is if create a `IDGenerator` that is force to receive the `ExecutionContext` on the method `generate` and can not accept in the constructor. So who call the method need to choice which `ExecutionContext` should be used otherwise it's not possible to inject the `ExecutionContext` on the implementation.

```scala
class FooIDGenerator() extends IDGenerator {
  def generate(implicit ec: ExecutionContext): Future[String]  = Future("foo")
}
```
VS
```scala
class FooIDGenerator(implicit ec: ExecutionContext) extends IDGenerator {
  def generate(): Future[String]  = Future("foo")
}
```

My suggestion is to review all place where need `ExecutionContext` and always prefers the constructor style.

joaoraf:
@rfranco 
I believe we may need to consider this carefully. In some cases, it makes perfect sense to fix an ExecutionContext, such as on the end of the call chain, where we usually block on I/O, such as when accessing a database, a remote service, etc. 
Slick, for example, uses its own dispatcher inside its DBIO Monad.
A remote access for a specific Auth Service may require a pool of threads of its own or, its use may be so rare that it is not worth the extra cost of setting up another ExecutionContext.
Between the layer above Silhouette and the bottom layer (whatever it is), I believe it is best to use the calling point execution context, since the code in the middle is just CPU bound (and not much intensive at that). 

What do you think?

Whatever you decide, I will try to make it.

Another path is to make it configurable somehow, allowing the user to choose (or inject) ExecutionContexts where such are needed.

akkie:
I see it similar to @joaoraf. An execution context should be defined context bound and not at initialization time(for classes that gets typically injected). For helper or internal classes that gets typically constructed inside the code, a constructor injected execution context makes sense.

I've searched through Play, Slick, ReactiveMongo and Akka, and all the projects prefers the execution context on the method instead of the constructor.

joaoraf:
Remember that you need a ExecutionContext whenever you manipulate a future using map, flatMap, etc. To create a Future you also need an ExecutionContext. What I believe should be avoided is the need to use the "scala.concurrent.ExecutionContext.global" execution context just because there is none when you need, for example, to map (or flatMap) the value of a future before returning to the caller.

rfranco:
@akkie @joaoraf agree with both usually i prefer to inject `ExecutionContext` on Service, Repository and DAO because these could have syncs calls or necessary different context as UserService (JDBC) and Cache (Redis).

So we really should review all the methods and see which need each approach by method or constructor.

akkie:
@joaoraf I meant methods that gets implemented by the user itself. Because the user can then define which execution context should be used. For methods that are already implemented by us the user should be able to pass the execution context.

Here are some examples for methods that gets implemented by the user:
[SecuredSettings.onNotAuthenticated](https://github.com/joaoraf/play-silhouette/blob/master/silhouette/app/com/mohiva/play/silhouette/api/SecuredSettings.scala#L41)
[SecuredSettings.onNotAuthorized](https://github.com/joaoraf/play-silhouette/blob/master/silhouette/app/com/mohiva/play/silhouette/api/SecuredSettings.scala#L52)
[Silhouette.onNotAuthenticated](https://github.com/joaoraf/play-silhouette/blob/master/silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala#L79)
[Silhouette.onNotAuthorized](https://github.com/joaoraf/play-silhouette/blob/master/silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala#L89)

DAOs, Services and repositories gets also implemented by the user. And I'm really unsure if it's the best approach to let the user handle the different methods over different execution contexts. So I'm with @rfranco here, to let the user inject the execution context if needed.

Anyway, I'm really unsure about the right approach. If I see it right then the current implementation provided by @joaoraf gives a user the possibility to define the execution context on controller level and then let the Silhouette depended code run in this execution context. A user is also able to define an other execution context in sub components of Silhouette.

If we implement only constructor injection of the execution context, then the user can only define the execution context for all controller instances and the dependent Silhouette classes once. As like with the other approach a user is able to define another execution context in sub components of Silhouette.

akkie:
Can we find any consensus here?

joaoraf:
Since Silhouette is intrinsically coupled to Play, we could use Play's ExecutionContext, which we can obtain through play.current or through the request (I believe) all way up to the calls to Services and Repositories (avoiding passing it through the call chain). The Services and Repositories then decide what they will use. They can use the Play's execution context also. We should just warn the user to avoid using the global execution context, using the Play one instead if they don't require a separate context for the Repository/Service.

As I said, whatever you decide I help implement. I am getting used to trace all these calls on the code ;-)

akkie:
I've created a topic is the [Scala-Users](https://groups.google.com/forum/#!topic/scala-user/Uncq6D4-O2g) mailing list with the title "Best strategy to pass the ExecutionContext for a library". Maybe this can help us to find the best solution.

joaoraf:
A very very nice discussion by the way. It seems this affects many people.

akkie:
@joaoraf Would it be OK for you if I create a new pull request. I find it hard to make decisions based on this pull request. I think I must rummage through the code to find the places which needs the `ExecutionContext`.

joaoraf:
No problem. Do you want some help? Have you decided the strategy?

akkie:
Fixed with #360

Issue #347, opened 2015-05-14
application.conf is out of date


aparamonov:
Hi!

Is there reason to have file 'application.conf' in this library?
If so, I think we should remove 'application.secret' configuration property or at least rename it to 'play.crypto.secret'. This will fix warning in applications that use Play 2.4.0: [[33mwarn[0m] application in application-akka.actor.default-dispatcher-3 Req: - application.conf @ jar:file:/Users/aparamonov/.ivy2/cache/com.mohiva/play-silhouette_2.11/jars/play-silhouette_2.11-3.0.0-SNAPSHOT.jar!/application.conf: 5: application.secret is deprecated, use play.crypto.secret instead

Also, I see another one warning: [[33mwarn[0m] application in main Req: - Logger configuration in conf files is deprecated and has no effect. Use a logback configuration file instead.
So, we should get rid of logging configuration properties in the file too.

Thanks,
Alexey

akkie:
Fixed with #349

PR #349, opened 2015-05-15
Fixes issue #347

joaoraf:
is the proper name for a default configuration file
provided by a library, according to the HOCON
convention.

Removed logging properties from the configuration.

Logging should be set by the main application, not the
library.

Fixes issue #347.


akkie:
Thanks @joaoraf! The old version of Play had a lot of logging issues, so we had to configure the logging level for Travis. Otherwise it has produced a lot of noise in the Travis build. It seems that this issues are gone now.

Issue #346, opened 2015-05-14
Is it possible to store session customized


yuikns:
I need to deploy several play servers, and I wish them share one jwt token system. Additionally, sometimes I need to restart the server to apply my update, it may cause users' token expired too quick.

Is it possible that I  store these tokens into redis/mongodb ?

Any help are appreciated.

akkie:
You can configure the `JWTAuthenticatorService` to use a `AuthenticatorDAO` implementation. In this DAO you can then implement the logic to store the token in a backing store.

PR #353, opened 2015-05-16
Add Tag Manifest for CacheLayer

yuikns:
As is said in #346 ,  I wish to manage the session myself and I can avoid the sessions' lost when I restart the server.

Another solution is configure the `JWTAuthenticatorService` and use a customized `CacheLayer` implementation.

Store the sessions into redis needs to serialize&unzerialize the value between T and String. Unzerialization need to get Manifest information.

I wish this pull request will be approved and someone like me may implement the `CacheLayer` more grace.

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/api/util/CacheLayer.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/api/util/CacheLayer.scala
@@ -31,7 +31,7 @@ trait CacheLayer {
    * @tparam T The type of the object to return.
    * @return The found value or None if no value could be found.
    */
-  def find[T: ClassTag](key: String): Future[Option[T]]
+  def find[T: ClassTag: Manifest](key: String): Future[Option[T]]
 
   /**
    * Save a value in cache.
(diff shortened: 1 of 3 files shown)


akkie:
The methods or class definitions you have changed, do have already the `ClassTag` which is entirely sufficient to serialize/unserialze your models. 

**Note:** The `Manifest` will be deprecated in favor of `TypeTags` and `ClassTags` 

http://docs.scala-lang.org/overviews/reflection/typetags-manifests.html
> In Scala 2.10, scala.reflect.ClassManifests are deprecated, and it is planned to deprecate scala.reflect.Manifest in favor of TypeTags and ClassTags in an upcoming point release. Thus, it is advisable to migrate any Manifest-based APIs to use Tags.

yuikns:
@akkie  , Thanks for your reply.
I found a related solution. [Convert a TypeTag to a Manifest](http://stackoverflow.com/questions/23383814/is-it-possible-to-convert-a-typetag-to-a-manifest )

PR #355, opened 2015-05-20
Add override modifier to methods

rfranco:
it helps to get a compiler error if a method signature is change.

```bash
[error] class Foo {
[error]       ^
[error] method 'foo' overrides nothing
[error]   override def foo() = {
[error]                ^
```


akkie:
Please could you rename the commit message into "Add override modifier to methods"

rfranco:
done

akkie:
:thumbsup:

PR #360, opened 2015-05-25
Alter the library methods to propagate an implicit ExecutionContext

akkie:
For the old discussion please see #342.

I have decided to separate the execution contexts by the relation of the single tasks. Normally you use a special execution context to separate a group of same tasks to let it run on dedicated thread pool. So for Silhouette we have the following different tasks:

* external HTTP requests (Gravatar requests, social provider requests)
* IO operations (database operations, accessing the file system (dev/random))
* internal operations (authenticator related actions or asynchronous callback execution)

All external HTTP requests can now use a single execution context which will be propagated through the `HTTPLayer` implementation. The DAO, service, repository and generator traits will not propagate an execution context through its methods. The execution context should be handled by the user through constructor injection. The `DelegableAuthInfoRepository` or the `SecureRandomIDGenerator` follows this pattern. For the Silhouette internal operations I've decided to propagate the execution context through the `Environment` as suggested by @rfranco.

What do you think @joaoraf and @rfranco?

### DIFF
--- a/silhouette-testkit/app/com/mohiva/play/silhouette/test/Fakes.scala
+++ b/silhouette-testkit/app/com/mohiva/play/silhouette/test/Fakes.scala
@@ -23,10 +23,11 @@ import com.mohiva.play.silhouette.api.util.Clock
 import com.mohiva.play.silhouette.impl.authenticators._
 import com.mohiva.play.silhouette.impl.daos.AuthenticatorDAO
 import com.mohiva.play.silhouette.impl.util.{ DefaultFingerprintGenerator, SecureRandomIDGenerator }
+import play.api.libs.concurrent.Execution.Implicits._
 import play.api.mvc.RequestHeader
 
 import scala.collection.mutable
-import scala.concurrent.Future
+import scala.concurrent.{ ExecutionContext, Future }
 import scala.reflect.runtime.universe._
 
 /**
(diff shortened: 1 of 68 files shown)


rfranco:
@akkie i'm with doubt about `AuthenticatorService` maybe that should extends `ExecutionContextProvider`.
Anyway it looks good to go! :+1:

akkie:
The idea with the `AuthenticatorService` was that it should be executed with the execution context passed to the `Environment`. In the Silhouette trait this may be true, but in the controller a user is able to pass a different execution context. I'm unsure about the right approach here. Should all authenticator methods use the same execution context?

joaoraf:
It depends. If some authentication method, say oauth2 with Google,
represents for example 90% of the cases, you might want to isolate it from
the others.

Em Ter, 26 de mai de 2015 06:08, Christian Kaps <notifications@github.com>
escreveu:

akkie:
@joaoraf I think you have misinterpreted my question. An authenticator has nothing to do with the provider.

joaoraf:
Sorry, I had just woken up! Please disregard it.

Em Ter, 26 de mai de 2015 10:46, Christian Kaps <notifications@github.com>
escreveu:

akkie:
I've updated the `AuthenticatorService`, `BasicAuthProvider` and `CredentialsProvider` to use the `ExecutionContextProvider`.

akkie:
@joaoraf Do you have any suggestions. Otherwise I would merge it.

joaoraf:
Nope. Please go ahead. Can't wait to have this settled down to stop using
my fork.
Again, thanks to all contributors for the very useful library.

Em Ter, 26 de mai de 2015 15:50, Christian Kaps <notifications@github.com>
escreveu:

akkie:
Thanks @rfranco and @joaoraf. I really appreciate your help on this issue. :+1:

PR #362, opened 2015-05-30
Bumping Play version to 2.4.0 final.

cemcatik:


akkie:
Thanks

cemcatik:
Thanks.. When can we expect new snapshot artifacts with this change?

akkie:
@cemcatik Snapshot is out

cemcatik:
@akkie Awesome thanks!

PR #374, opened 2015-06-11
Twitter Provider: add email support

Lykathia:
Earlier in 2015 twitter added the ability for applications to retrieve user emails. Applications have to be specifically whitelisted in order to do this, and an additional token has to be sent along with the request to verify_credentials in order for the email to be returned.


akkie:
Nice! Could you please squash the both commits into a single one?

Lykathia:
Had them separated as they are two functionally different things. Are you sure you would rather see them squashed together?

akkie:
Yes please! This makes it easier to review and also makes reverting the code easier in case of any build breakages.

Lykathia:
done 'n done

akkie:
Thanks :+1:

PR #382, opened 2015-06-15
Avoid requiring a started Play! application just to create a default CookieAuthenticatorSettings object

bwbecker:
Many of my tests would use default CookieAuthenticatorSettings object.  But that throws an exception if there is no started application in scope.  This change causes secureCookie to default to false in that case.

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/api/Silhouette.scala
@@ -267,12 +267,12 @@ trait Silhouette[I <: Identity, A <: Authenticator] extends Controller with Logg
       val auth = env.authenticatorService.touch(authenticator)
       block(auth.extract).flatMap {
         case hr @ HandlerResult(pr: Authenticator.Discard, _) =>
-          env.authenticatorService.discard(authenticator, Future.successful(pr)).map(pr => hr.copy(pr))
+          env.authenticatorService.discard(authenticator, pr).map(pr => hr.copy(pr))
         case hr @ HandlerResult(pr: Authenticator.Renew, _) =>
-          env.authenticatorService.renew(authenticator, Future.successful(pr)).map(pr => hr.copy(pr))
+          env.authenticatorService.renew(authenticator, pr).map(pr => hr.copy(pr))
         case hr @ HandlerResult(pr, _) => auth match {
           // Authenticator was touched so we update the authenticator and maybe the result
-          case Left(a) => env.authenticatorService.update(a, Future.successful(pr)).map(pr => hr.copy(pr))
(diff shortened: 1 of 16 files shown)


akkie:
Hi Byron,

with the next Play release in mind, I think we can replace all `Play.isProd` calls with a Boolean value, because, this functionality relies on global state which gets completely removed from Play in the future. In Play the default value for the Cookie `secure` property is set to false, so I think we can set it also to false and add a warning to the documentation.

What do you think?

bwbecker:
In general, I prefer to default to the more secure setting.  But it probably makes sense to follow the lead from Play itself.

Thanks for checking.  I have not objections to this approach rather than the one I suggested in my PR. 

Byron

> On Jun 16, 2015, at 2:48 PM, Christian Kaps <notifications@github.com> wrote:
> 
> Hi Byron,
> 
> with the next Play release in mind, I think we can replace all Play.isProd calls with a Boolean value, because, this functionality relies on global state which gets completely removed from Play in the future. In Play the default value for the Cookie secure property is set to false, so I think we can set it also to false and add a warning to the documentation.
> 
> What do you think?
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/mohiva/play-silhouette/pull/382#issuecomment-112526658>.
> 


---------------------------------------------
Byron Weber Becker
School of Computer Science
University of Waterloo, Waterloo, Ontario  N2L 3G1
Email: bwbecker@uwaterloo.ca
Phone:  519-888-4567 x34661
Fax:  519-888-4380 (include my email on cover page)

Advising FAQ:  www.cs.uwaterloo.ca/current/faq

akkie:
You are right, we should prefer the more secure setting by default. Would you create a pull request?

bwbecker:
Christian — I’d love to create a pull request, but I’m pushing hard to get ready for a major holiday (Europe!) and certainly would not be able to get to it until August at the earliest.  I haven’t yet done anything with Play 2.4, so would be uncertain about targeting it anyway.  Sorry.

Byron


> On Jun 16, 2015, at 3:07 PM, Christian Kaps <notifications@github.com> wrote:
> 
> You are right, we should prefer the more secure setting by default. Would you create a pull request?
> 
> —
> Reply to this email directly or view it on GitHub <https://github.com/mohiva/play-silhouette/pull/382#issuecomment-112531712>.
> 


---------------------------------------------
Byron Weber Becker
School of Computer Science
University of Waterloo, Waterloo, Ontario  N2L 3G1
Email: bwbecker@uwaterloo.ca
Phone:  519-888-4567 x34661
Fax:  519-888-4380 (include my email on cover page)

Advising FAQ:  www.cs.uwaterloo.ca/current/faq

akkie:
No problem! Have a nice trip! May I ask what countries do you travel to in Europe?

akkie:
Close this in favor of #384

Issue #386, opened 2015-06-24
Add get[T: ClassTag](id: String): Option[T] method to the SocialProviderRegistry


daron666:
Somtenies it will be usefull to get social profile without authorization only by token. But this code
```scala
val id = "facebook"
val profile = socialProviderRegistry.get(id) match {
    case Some(p) => p.retrieveProfile(OAuth2Info(accessToken = "token")
    case None => Future.failed(...)
}
```
returns error, cause retrieveProfile expects provider.A type instead of OAuth2Info.type.
Perhaps, method from the topic will resolve this problem. We will be able to write 
```scala
val id = "facebook"
val profile = socialProviderRegistry.get[OAuth2Provider](id) match {
    case Some(p) => p.retrieveProfile(OAuth2Info(accessToken = "token")
    case None => Future.failed(...)
}
```

akkie:
Fixed with #392

PR #387, opened 2015-06-24
fix mohiva/play-silhouette#386

daron666:


akkie:
I'm not entire happy with the naming of the method. Maybe we can find a solution with only one method for both cases. Let's investigate if this is possible.

daron666:
Yeah. No problem. I'm going to investigate this in nearest future/

rfranco:
the name could be `getAs`

akkie:
@rfranco Great idea! I think the new interface should look like this:
```scala
def getAs[T: ClassTag]: Option[T]

def getByID(id: String): Option[SocialProvider]

def getByIDAs[T <: SocialProvider: ClassTag](id: String): Option[T]

def getSeqAs[T: ClassTag]: Seq[T]
```

Or we use a single method for both ID cases:
```scala
def get[T: ClassTag]: Option[T]

def get[T <: SocialProvider: ClassTag](id: String): Option[T]

def getSeq[T: ClassTag]: Seq[T]
```

But then the user must always give the type to return:
```scala
socialProviderRegistry.get[SocialProvider](provider)
```

What do you think?

daron666:
@akki I like new interface.

rfranco:
@akkie i prefer the second option that give a clean interface

akkie:
I prefer also the second. I'll create a pull request.

akkie:
Fixed with #392

PR #411, opened 2015-09-23
Make actions injectable

akkie:
This is a work in progress pull request to get feedback for the next version.

These are the things I have changed:
- Removed Play `Messages` instance from `Event`, `Authorization` and `ErrorHandler` types. The `I18nSupport` trait should be used instead, to get the `Messages` instance for the current request
- Removed `Silhouette` trait in favour of injectable actions (this is now the default method from Play 2.5 on)
  - Every `SecuredAction` can now override the global(default injected) error handler if needed
  - A controller is not bound to a single `Authenticator` anymore
- Remove `SecuredErrorHandler` in favour of injectable error handler

@rfranco Any suggestions? I think the omitting of the `Messages` instance is something you'll like.


rfranco:
@akkie i did a quick review and it looks great, i'll do a better review and add some comment later.

akkie:
I have rewritten the code to provide Play DI modules for the actions. I think this is the better implementation because a user doesn't have to bind the actions by itself for every environment combination. As a side effect the user must now pass the environment to the action.

rfranco:
@akkie The `Actions` and `Env` look so complicate now, because i need to inject enviroment and pass that for every actions. I'm not sure if it's the better way to go.

Maybe could try to inject only the `Enviroment` and use likely:

```scala 
  class SecuredController @Inject() (
    private val env: Environment[FakeIdentity, FakeAuthenticator])
    extends Controller {

    def defaultAction = env.SecuredAction { implicit request =>
      render {
        case Accepts.Json() => Ok(Json.obj("result" -> "full.access"))
        case Accepts.Html() => Ok("full.access")
      }
    }
}
```

What do you think?

akkie:
It looks better, but then the user must inject the actions into the environment, which would make the test code very verbose. Especially if you use the `FakeEnvironment`. I'm not sure which is the best approach.  But it should not be a problem if we write `env.SecuredAction` or `securedAction(env)`.

The main reason to provide `Env` was also to reduce boilerplate code. All over the code you have to write `FakeIdentity, FakeAuthenticator` to describe your environment. If you define the `Env` once you can use a more shorter and descriptive term. It is also less error prone because you cannot accidentally use another authenticator type. It's also easier to refactor. And if you look through the teskit specs, then you can see that you avoid to write a lot of boilerplate code.

```scala
implicit val env = FakeEnvironment[CookieEnv](Seq(loginInfo -> identity))
val request = FakeRequest().withAuthenticator(loginInfo)
```

instead of

```scala
implicit val env = FakeEnvironment[FakeIdentity, CookieAuthenticator](Seq(loginInfo -> identity))
val request = FakeRequest().withAuthenticator[CookieAuthenticator](loginInfo)
```

akkie:
@rfranco Maybe we could introduce a wrapper class which contains all the injectable components? Something like an `ActionStack` or `EnvironmentStack`. The problem here is that the user must also bind the instance by hand because of the raw type. But we could avoid the test verbosity that we would introduce if we let the `Environment` provide all the components.

What do you think?

akkie:
```scala
class SecuredController @Inject() (
  silhouette: Silhouette[FakeEnv])
  extends Controller {

  def defaultAction = silhouette.SecuredAction { implicit request =>
    render {
      case Accepts.Json() => Ok(Json.obj("result" -> "full.access"))
      case Accepts.Html() => Ok("full.access")
    }
  }
}
```

And the user must only bind additionally the `Environment`.

```scala
bind[Silhouette[FakeEnv]].to[SilhouetteProvider[FakeEnv]]
```

What do you think @rfranco ?

rfranco:
@akkie sorry for my delay to answer, i was traveling

for me it's perfect now. it's simple to understand and use.

akkie:
No worries! I think also it's perfect now.

PR #390, opened 2015-11-12
Use request extractors to find authenticator values in other parts of the request

akkie:
Sometimes it may be necessary to send the authenticator specific values in other parts of the request. If we use a request extractor in the `retrieve` method of the authenticator service, then it may be possible to send a bearer token as example in the query string instead of the header.

See: http://silhouette.mohiva.com/v3.0/discuss/55950f4935f567170094304c

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
@@ -19,7 +19,7 @@
  */
 package com.mohiva.play.silhouette.api.services
 
-import com.mohiva.play.silhouette.api.util.ExecutionContextProvider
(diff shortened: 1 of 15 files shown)


seyuf:
Hi guys,
I'd like to know when will the v 4.0.0 be available?
I need badly this feature, i tried to use the 4.0.0-snapshot from sonatype.
Unfortunately, it is not compatible with my project as is, Silhouette is no longer extended from the controller etc... 
May be we could have some sample displaying the new usage of the module.
Anyway thanks for this awesome work :+1:

akkie:
The final 4.0 will be available with Play 2.5 at the beginning of the next year. A migration guide can be found in the [documentation](http://silhouette.mohiva.com/v4.0/docs/migration-guide).

PR #427, opened 2015-11-20
Cas provider 3.0.0

arjunsol:
Hi,

There is work in this pull to add a new provider for jasig CAS.

Unfortunately it was initially built on v2 and then merged with the v3 tag. The pull includes all that merging and I doubt will be compatible with master.

Can you let me know how you'd like to proceed?


akkie:
This pull request cannot be merged because of the conflicts. So either you cherry pick [this commit](https://github.com/arjunsol/play-silhouette/commit/b21fd6be958f9533e807bbce94282090435f887c) and apply it on top of master or you copy the files you've changed and create a new pull request only with this files.

Anyway, nice to see CAS support in Silhouette :+1:

arjunsol:
I think I'll go with the second option, though I believe as things stand in master the silhouette seed is incompatible?

akkie:
Yes the seed templates are not adapted to the current master branch. I will adapt it if Silhouette 4 is in the release phase. This is with the release of Play 2.5 in January. Maybe earlier.

PR #433, opened 2015-12-12
Fix overriding settings method for Providers.

### LINKED ISSUE #432: Fix overriding settings method for Providers
SocialProvider trait exposes withSettings method to override settings dependency. Current implementation only overrides first level settings dependency, but it doesn't do the same with settings wrapped by internal services.

felipefzdz:
This PR fixes #432. Some notes: 

There are three flavours of SocialProvider: OAuth1, OpenID and OAuth2. OAuth1 and OpenID related services only hold references into its respective first level settings objects (OAuth1Settings and OpenIDSettings). That makes `withSettings` API really appropriate. However OAuth2 related services (e.g CookieStateProvider) have different settings objects: CookieStateSettings vs OAuth2Settings. That means that this API could be misleading for OAuth2 scenarios, it seems that we're updating every settings value related with that OAuth2 provider but actually we're only updating OAuth2Settings instances. That's out of the scope for this fix, as it would probably involve a new API or updating the existing one, but I'd like to note it.

About OAuth1 and OpenID, one of the design decisions was exposing similar `withSettings` for their services (OAuth1Service and OpenIDService). That allow us to keep the extensibility of the current design and being able at the same time to access to internal dependencies like `client` in `PlayOpenIDService`. Unit tests for those new functionalities have been provided.

This PR breaks isolation in unit tests like `LinkedInProviderSpec`. Being able to check that some concrete setting was updated in the related services seemed more useful than verifying that some mock was called. If that's breaking the philosophy of these specs, please let me know and I'll update the PR.

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth1Provider.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth1Provider.scala
@@ -131,6 +131,11 @@ object OAuth1Provider {
 trait OAuth1Service {
 
   /**
+   * The type of the concrete implementation of this abstract type.
+   */
+  type Self <: OAuth1Service
+
+  /**
    * Indicates if the service uses the safer 1.0a specification which addresses the session fixation attack
    * identified in the OAuth Core 1.0 specification.
    *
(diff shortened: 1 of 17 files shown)


akkie:
Thanks for working on this! I think a better approach for the tests would be to use [argument capture](http://blog.themillhousegroup.com/2013/11/argument-capture-with-scala-specs2.html) to check if the `withSettings` method on the related service` was called with the appropriate argument. Then you could use the provider instance provided in every test.

What do you think?

felipefzdz:
You're welcome :) I'm happy to use the mocking approach. The service needs to be stubbed and not mocked, as it needs to return some settings object, otherwise it crashes with NPE in further initialisation. It seems that using Argument Captor is not recommended for stubs as you can read [here](http://docs.mockito.googlecode.com/hg/org/mockito/ArgumentCaptor.html). As result of that I'm using a simple verifying step matching against the provided overriding settings function.

```there was one(oAuthService).withSettings(overrideSettingsFunction)```

I forced pushed the changes with this improvement, hope you don't mind.

akkie:
Good work! Thanks!

PR #436, opened 2015-12-16
issue#435 * copy customClaims when renewing JWTToken

mizerlou:
this should fix issue#435.  is there anything else that should be copied?  idleTimeout, maybe?


akkie:
Great! Could you please squash the both commits into one.

https://docs.openshift.org/origin-m4/oo_contributors_guide.html#combining-multiple-commits

akkie:
Thanks!

PR #439, opened 2016-02-09
Fix broken tests introduced in ad7bcdc5

lucamilanesio:
Since the introduction of API URL override in ad7bcdc5 the build was
inevitably broken because of a wrong Mockito validate statement. 
Travis did not detect the breakage because of the missing "-e" option
on the bash execution, thus the final "echo" was shallowing the exit
code of the test.

This change fixes the build ci script *AND* the Mockito validation
statement.

### DIFF
--- a/scripts/ci-build
+++ b/scripts/ci-build
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
 #
 # Builds the project in the continuous integration environment.
 #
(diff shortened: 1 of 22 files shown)


lucamilanesio:
@akkie this fixes the broken build https://travis-ci.org/mohiva/play-silhouette/builds/107967523

akkie:
Thanks. There are some other failing test I'll fix later.

akkie:
Sorry, I cannot merge the pull request because of the failed build. Please could you fix the other issues too? It seems that the default timeout from one second isn't enough for some of the tests.

lucamilanesio:
@akkie I believe this is a more general issue with the way that waits are managed in the tests. Instead of using a global configuration, possibly overridable by a system property, they are using the .wait method that put a hard failure after a 1 sec timeout. When you are unlucky and Travis is much slower than usual .... random tests may fail.

lucamilanesio:
@akkie most broken test should be fixed now. Bearing in mind that there are a tons of other "wait 1 sec" across the Specs ... and depending on how quick is Travis, they may fail again ;-(

akkie:
@lucamilanesio Nice solution! I'll look over the code and try to fix it in a similar way. Thanks!

PR #440, opened 2016-02-10
Support for Auth0 authentication provider

lucamilanesio:
Introduce the support of Auth0.com, a popular authentication service
which exposes a OAuth 2.0 provider interface.

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0Provider.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0Provider.scala
@@ -1,0 +1,188 @@
+/**
+ * Copyright 2015 Mohiva Organisation (license at mohiva dot com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.mohiva.play.silhouette.impl.providers.oauth2
+
+import com.mohiva.play.silhouette.api.LoginInfo
+import com.mohiva.play.silhouette.api.util.HTTPLayer
+import com.mohiva.play.silhouette.impl.exceptions.ProfileRetrievalException
+import com.mohiva.play.silhouette.impl.providers._
+import com.mohiva.play.silhouette.impl.providers.oauth2.Auth0Provider._
+import play.api.libs.json.{ JsString, JsValue }
+import play.api.mvc.RequestHeader
+import play.mvc.Http
+
+import scala.concurrent.Future
+
+/**
+ * Base Auth0 OAuth2 Provider.
+ *
+ * OAuth Provider configuration in silhouette.conf must indicate:
+ *
+ *   # Auth0 Service URLs
+ *   auth0.authorizationURL="https://mydomain.eu.auth0.com/authorize"
+ *   auth0.accessTokenURL="https://mydomain.eu.auth0.com/oauth/token"
+ *   auth0.apiURL="https://mydomain.eu.auth0.com/userinfo"
+ *
+ *   # Application URL and credentials
+ *   auth0.redirectURL="http://localhost:9000/authenticate/auth0"
+ *   auth0.clientID=myoauthclientid
+ *   auth0.clientSecret=myoauthclientsecret
+ *
+ *   # Auth0 user's profile information requested
+ *   auth0.scope="openid name email picture"
+ *
+ * See http://auth0.com for more information on the Auth0 Auth 2.0 Provider and Service.
+ */
+trait BaseAuth0Provider extends OAuth2Provider {
+
+  /**
+   * The content type to parse a profile from.
+   */
+  override type Content = JsValue
+
+  /**
+   * The provider ID.
+   */
+  override val id = ID
+
+  /**
+   * Defines the URLs that are needed to retrieve the profile data.
+   */
+  override protected val urls = Map("api" -> settings.apiURL.getOrElse(API))
+
+  /**
+   * Builds the social profile.
+   *
+   * @param authInfo The auth info received from the provider.
+   * @return On success the build social profile, otherwise a failure.
+   */
+  override protected def buildProfile(authInfo: OAuth2Info): Future[Profile] = {
+    val request = httpLayer.url(urls("api"))
+    val requestWithHeader = request.withHeaders(("Authorization", s"Bearer ${authInfo.accessToken}"))
+
+    val httpResponse = requestWithHeader.get()
+    httpResponse.flatMap { response =>
+      response.status match {
+        case Http.Status.OK =>
+          profileParser.parse(response.json, authInfo)
+
+        case _ => {
+          val json = response.json
+          (json \ "error").asOpt[JsString] match {
+            case Some(error) =>
+              val errorDescription = (json \ "error_description")
+              throw new ProfileRetrievalException(SpecifiedProfileError.format(id, errorDescription, error, response.status))
+            case _ =>
+              throw new ProfileRetrievalException(GenericHttpStatusProfileError.format(id, response.status))
+          }
+        }
+      }
+    }
+  }
+
+  /**
+   * Gets the access token.
+   *
+   * @param code    The access code.
+   * @param request The current request.
+   * @return The info containing the access token.
+   */
+  override protected def getAccessToken(code: String)(implicit request: RequestHeader): Future[OAuth2Info] = {
+    request.getQueryString("token_type") match {
+      case Some("bearer") => Future(OAuth2Info(code))
+      case _ => super.getAccessToken(code)
+    }
+  }
+}
+
+/**
+ * The profile parser for the common social profile.
+ */
+class Auth0ProfileParser extends SocialProfileParser[JsValue, CommonSocialProfile, OAuth2Info] {
+
+  /**
+   * Parses the social profile.
+   *
+   * @param json The content returned from the provider.
+   * @return The social profile from given result.
+   */
+  override def parse(json: JsValue, authInfo: OAuth2Info): Future[CommonSocialProfile] = Future.successful {
+    val userID = (json \ "user_id").as[String]
+    val avatarURL = (json \ "picture").asOpt[String]
+    val email = (json \ "email").asOpt[String]
+
+    CommonSocialProfile(
+      loginInfo = LoginInfo(ID, userID),
+      avatarURL = avatarURL,
+      email = email)
+  }
+}
+
+/**
+ * The Auth0 OAuth2 Provider.
+ *
+ * @param httpLayer The HTTP layer implementation.
+ * @param stateProvider The state provider implementation.
+ * @param settings The provider settings.
+ */
+class Auth0Provider(
+  protected val httpLayer: HTTPLayer,
+  protected val stateProvider: OAuth2StateProvider,
+  val settings: OAuth2Settings)
+  extends BaseAuth0Provider with CommonSocialProfileBuilder {
+
+  /**
+   * The type of this class.
+   */
+  override type Self = Auth0Provider
+
+  /**
+   * The profile parser implementation.
+   */
+  override val profileParser = new Auth0ProfileParser
+
+  /**
+   * Gets a provider initialized with a new settings object.
+   *
+   * @param f A function which gets the settings passed and returns different settings.
+   * @return An instance of the provider initialized with new settings.
+   */
+  override def withSettings(f: (Settings) => Settings) = new Auth0Provider(httpLayer, stateProvider, f(settings))
+}
+
+/**
+ * The companion object.
+ */
+object Auth0Provider {
+
+  /**
+   * The error messages.
+   */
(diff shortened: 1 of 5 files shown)


silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialProvider.scala
```diff
@@ -165,6 +165,14 @@ trait SocialProfile {
   def loginInfo: LoginInfo
 }
 
+trait IdentifiedSocialProfile extends SocialProfile {
```
akkie: Why do you have implemented the `IdentifiedSocialProfile` trait? This isn't necessary and it forces every social profile to implement the fields and it breaks all the application which doesn't use the `CommonSocialProfileBuilder`.

lucamilanesio: @akkie I needed to subclass CommonSocialProfile ... but I couldn't as it is a case class and not a trait :-( The only choice was then introduce a trait that CommonSocialProfiles implements (and thus the IdentifiedSocialProfile) and have the new Auth0SocialProfile implementing it. Do you see a better way to get the same result?

akkie: It should be the same as described in the [documentation](http://silhouette.mohiva.com/docs/providers#section-write-a-custom-social-profile-builder). Or is there a problem I can't see?

akkie:
I'm also not sure if we should implement a `Auth0Provider` which is compatible with `CommonSocialProfile` and on top of this a `ExtendedAuth0Provider`with a `ExtendedAuth0SocialProfile`. The problem is that the current provider isn't compatible with the rest of the providers and therefore cannot be used as drop-in replacement in the seed templates. The more experienced developer can then write the additional code in the seed template to support the extended version. What do you think?

lucamilanesio:
I could have just implemented a CustomSocialProfile as described in the doc, however the SocialProfile defines so little that it would not been used as drop-in replacements for Facebook or Twitter :-( The idea of having two (one compatible and one extended) would require quite some copy&paste, which isn't nice at all :-(

akkie:
> I could have just implemented a CustomSocialProfile as described in the doc, however the SocialProfile defines so little that it would not been used as drop-in replacements for Facebook or Twitter 

Yes, the user must extend the [SocialAuthController](https://github.com/mohiva/play-silhouette-seed/blob/master/app/controllers/SocialAuthController.scala#L43) and handle the custom provider in a separate way.

```scala
case Some(p: SocialProvider with Auth0ProfileBuilder) =>
```

If we merge your current implementation, then the user can also not access some of the fields you have implemented in the `Auth0SocialProfile`. The user must use the same pattern match as described above.

> The idea of having two (one compatible and one extended) would require quite some copy&paste, which isn't nice at all :-(

No, it doesn't require copy&paste, because it can reuse the other provider. It's the same as described in the [docs](http://silhouette.mohiva.com/docs/providers#section-write-a-custom-social-profile-builder). The `CustomFacebookProvider` uses the parts from the `FacebookProvider`.

But, normally Silhouette should only provide compatible providers. If someone requires more information then he/she should extend the provider as described in the docs.

I see the problem, but we need a consensuses between the providers.

lucamilanesio:
@akkie ok makes sense. I would then rework this PR using the Auth0 + CustomAuth0 provider as in the Facebook example.

silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0Provider.scala
```diff
@@ -0,0 +1,198 @@
...
+  override def withSettings(f: (Settings) => Settings) = new Auth0Provider(httpLayer, stateProvider, f(settings))
+
+  /**
+   * Gets the access token.
+   *
+   * @param code    The access code.
+   * @param request The current request.
+   * @return The info containing the access token.
+   */
+  override protected def getAccessToken(code: String)(implicit request: RequestHeader): Future[OAuth2Info] = {
```
akkie: This method should be placed in the `BaseAuth0Provider` so that it can be reused.

silhouette/test/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0ProviderSpec.scala
```diff
@@ -0,0 +1,175 @@
...
+import test.Helper
+
+import scala.concurrent.Future
+
+/**
+ * Test case for the [[Auth0Provider]] class.
+ */
+class Auth0ProviderSpec extends OAuth2ProviderSpec {
+
+  val AUTH0_ERROR_JSON: String = "providers/custom/auth0.error.json"
```
akkie: Constants in Scala shall be written in camel-case and they should be placed in a companion object. The type declaration is also not needed.

http://docs.scala-lang.org/style/naming-conventions.html

silhouette/test/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0ProviderSpec.scala
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  "The `retrieveProfile` method" should {
+    "fail with ProfileRetrievalException if API returns error" in new WithApplication with Context {
+      val requestHolder = mock[WSRequest]
+      val response = mock[WSResponse]
+      private val statusCode: Int = Http.Status.UNAUTHORIZED
+      response.status returns statusCode
+      requestHolder.get() returns Future.successful(response)
+      val oauth2Info = oAuthInfo.as[OAuth2Info]
```
akkie: Could you please define the variables above the mock expectation definitions. This would make the code more readable, it's divided into logical blocks. In the other tests too.

silhouette/test/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0ProviderSpec.scala
```diff
@@ -0,0 +1,175 @@
...
+        case authInfo => authInfo must be equalTo oAuthInfo.as[OAuth2Info]
+      }
+    }
+  }
+
+  "The `retrieveProfile` method" should {
+    "fail with ProfileRetrievalException if API returns error" in new WithApplication with Context {
+      val requestHolder = mock[WSRequest]
+      val response = mock[WSResponse]
+      private val statusCode: Int = Http.Status.UNAUTHORIZED
```
akkie: Why private? I think we should not mix different variable definition styles because of the readability.

silhouette/test/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0ProviderSpec.scala
```diff
@@ -0,0 +1,175 @@
...
+    "fail with ProfileRetrievalException if an unexpected error occurred" in new WithApplication with Context {
+      val requestHolder = mock[WSRequest]
+      val response = mock[WSResponse]
+      response.json throws new RuntimeException("")
+      requestHolder.get() returns Future.successful(response)
+      val oauth2Info = oAuthInfo.as[OAuth2Info]
+      requestHolder.withHeaders(("Authorization", s"Bearer ${oauth2Info.accessToken}")) returns requestHolder
+      httpLayer.url(oAuthSettings.apiURL.get) returns requestHolder
+
+      failed[ProfileRetrievalException](provider.retrieveProfile(oAuthInfo.as[OAuth2Info])) {
```
akkie: The `oauth2Info` var could be reused here.

silhouette/test/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0ProviderSpec.scala
```diff
@@ -0,0 +1,175 @@
...
+      val response = mock[WSResponse]
+      response.status returns Http.Status.OK
+      private val userProfile: JsValue = Helper.loadJson(AUTH0_USER_PROFILE_JSON)
+      response.json returns userProfile
+      requestHolder.get() returns Future.successful(response)
+      val oauth2Info = oAuthInfo.as[OAuth2Info]
+      requestHolder.withHeaders(("Authorization", s"Bearer ${oauth2Info.accessToken}")) returns requestHolder
+      httpLayer.url(oAuthSettings.apiURL.get) returns requestHolder
+
+      profile(provider.retrieveProfile(oAuthInfo.as[OAuth2Info])) {
```
akkie: The `oauth2Info` var could be reused here.

silhouette/test/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0ProviderSpec.scala
```diff
@@ -0,0 +1,175 @@
...
+     * The OAuth2 info returned by Auth0.
+     */
+    override lazy val oAuthInfo = Helper.loadJson(AUTH0_SUCCESS_JSON)
+
+    /**
+     * The provider to test.
+     */
+    lazy val provider = new Auth0Provider(httpLayer, stateProvider, oAuthSettings)
+  }
+
```
akkie: The newline is not needed

silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0Provider.scala
```diff
@@ -0,0 +1,198 @@
...
+          profileParser.parse(response.json, authInfo)
+
+        case _ => {
+          val json = response.json
+          (json \ "error").asOpt[JsString] match {
+            case Some(error) =>
+              val errorDescription = (json \ "error_description")
+              throw new ProfileRetrievalException(SpecifiedProfileError.format(id, errorDescription, error, response.status))
+            case _ =>
+              throw new ProfileRetrievalException(s"Cannot get user profile - HTTP status code ${response.status}")
```
akkie: Could you please use a constant for the error message. So the exception case can be better tested.

silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0Provider.scala
```diff
@@ -0,0 +1,198 @@
...
+
+  /**
+   * The Facebook constants.
+   */
+  val ID = "auth0"
+
+  val API = "https://auth0.auth0.com/userinfo"
+}
+
+trait Auth0ProfileBuilder {
```
akkie: Doc comment is missing.

akkie:
@lucamilanesio that would be great :+1:  I've added some comments regarding some code style issues.

silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0Provider.scala
lucamilanesio: OK, done.

silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0Provider.scala
lucamilanesio: Makes sense, done.

silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0Provider.scala
lucamilanesio: Done.

silhouette/test/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0ProviderSpec.scala
lucamilanesio: I know, the types declarations are introduced by IntelliJ when you extract a constant ... hate it so much :-(
Done.

silhouette/test/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0ProviderSpec.scala
lucamilanesio: IntelliJ refactoring ... again ;-( I should just remove it from my system and use Sublime instead :-)
Done.

silhouette/test/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0ProviderSpec.scala
lucamilanesio: done.

silhouette/test/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0ProviderSpec.scala
lucamilanesio: Done and set in the Context.

silhouette/test/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0ProviderSpec.scala
lucamilanesio: As before, done.

silhouette/test/com/mohiva/play/silhouette/impl/providers/oauth2/Auth0ProviderSpec.scala
lucamilanesio: Removed.

lucamilanesio:
@akkie all changes done, build running .... hopefully we should be close now :-)

akkie:
Looks good! One last question. The Json contains a name and a nickname field. Should we not use one of these fields for the name?

lucamilanesio:
@akkie none of them can be associated to the CommonSocialProfile. Your login to Auth0 with e-mail and password, so e-mail is one we store in the silhouette's profile. Assuming that Auth0's name or nickname are part of the user's first or last or full name would be a mistake.

lucamilanesio:
Any more change / amendment required? Good to merge?

akkie:
No, it looks good. Thanks for your work!

Issue #446, opened 2016-03-08
Build against Play 2.5?


lunaryorn:
Silhouette fails to build against Play 2.5, giving these failures after changing the Play version in `plugins.sbt` to 2.5.0:

```
[error] [...]/play-silhouette/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala:46: too many arguments for constructor Result: (header: play.api.mvc.ResponseHeader, body: play.api.http.HttpEntity)play.api.mvc.Result
[error] class AuthenticatorResult(result: Result) extends Result(result.header, result.body, result.connection) {
[error]                                                   ^
[error] [...]/play-silhouette/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala:59: not found: value HttpConnection
[error]     connection: HttpConnection.Connection = this.connection) = {
[error]                 ^
[error] [...]/play-silhouette/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala:61: too many arguments for method copy: (header: play.api.mvc.ResponseHeader, body: play.api.http.HttpEntity)play.api.mvc.Result
[error]     AuthenticatorResult(super.copy(header, body, connection))
[error]                                   ^
[error] [...]/play-silhouette/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala:58: type mismatch;
[error]  found   : play.api.http.HttpEntity
[error]  required: play.api.libs.iteratee.Enumerator[Array[Byte]]
[error]     body: Enumerator[Array[Byte]] = this.body,
[error]                                          ^
[error] [...]/play-silhouette/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala:59: value connection is not a member of com.mohiva.play.silhouette.api.services.AuthenticatorResult
[error]     connection: HttpConnection.Connection = this.connection) = {
[error]                                                  ^
[error] 5 errors found
[error] (play-silhouette/compile:compileIncremental) Compilation failed
[error] Total time: 9 s, completed 08.03.2016 11:48:06
```

These errors seem to be related to [Play 2.5 using akka-streams](https://www.playframework.com/documentation/2.5.x/StreamsMigration25).

As things stand this prevents us from using Silhouette with Play 2.5, because Play 2.4 builds are incompatible with Play 2.5.

akkie:
I work on the build and I plan to release it this afternoon.

akkie:
[...]

PR #447, opened 2016-03-08
Some changes for Play 2.5

JamesSullivan:
Some minor changes to get this to build with Play 2.5. I did not address the crypto related changes in Play 2.5 so I had to turn off -Xfatal-warnings to get it to compile with those warnings. Also there are some tests failing around ErrorHandlerSpec but I believe those already existed and have nothing to do with these code changes.


akkie:
@JamesSullivan Thanks, I've a running build with all changes already at home on my disk. This build fixes also the `ErrorHandlerSpec` issues which are related to https://www.playframework.com/documentation/2.5.x/Migration25#Content-Type-charset-changes.

I'll provide a BETA version in a few hours when I'm home from work.

You can follow #446

PR #454, opened 2016-03-26
Include logRounds in BCryptPasswordHasher.id

alexmojaki:
My understanding of `CredentialsProvider` is that it is to be set up something like this:

```scala
val passwordHasher = new BCryptPasswordHasher(10)
new CredentialsProvider(authInfoRepository, passwordHasher, List(passwordHasher))
```

but then someday someone may want to increase the number of rounds, and so they'll do this:

```scala
val passwordHasher = new BCryptPasswordHasher(12)
new CredentialsProvider(authInfoRepository, passwordHasher, List(passwordHasher, new BCryptPasswordHasher(10)))
```

This can only work correctly if the hashers have distinct IDs, which they don't.
Perhaps https://github.com/mohiva/play-silhouette/blob/master/silhouette/test/com/mohiva/play/silhouette/impl/providers/CredentialsProviderSpec.scala should include a test with actual hashers and not just mocks to prove this.

### DIFF
--- a/silhouette-password-bcrypt/src/main/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasher.scala
+++ b/silhouette-password-bcrypt/src/main/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasher.scala
@@ -36,7 +36,7 @@ class BCryptPasswordHasher(logRounds: Int = 10) extends PasswordHasher {
    *
    * @return The ID of the hasher.
    */
-  override def id = ID
+  override def id = ID + logRounds
 
   /**
    * Hashes a password.
--- a/silhouette/app/com/mohiva/play/silhouette/impl/providers/CredentialsProvider.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/providers/CredentialsProvider.scala
@@ -49,6 +49,9 @@ class CredentialsProvider @Inject() (
   passwordHasher: PasswordHasher,
   passwordHasherList: Seq[PasswordHasher])(implicit val executionContext: ExecutionContext)
   extends Provider with ExecutionContextProvider {
+    
+  require(passwordHasherList.map(_.id).toSet.size == passwordHasherList.size,
+    "Every PasswordHasher must have a unique ID")
 
   /**
    * Gets the provider ID.


akkie:
A bcrypt password hash includes the rounds the hash wars created with. So there is no need to save the number of rounds otherwise.

If we test the credentials provider with a real password hasher instead of a mock, then this would violate against common test practices. There is also no real password hasher to test the credentials provider with. Do you have seen that the bcrypt password hasher is in a different package?

alexmojaki:
Ah, I didn't know that the hash included the number of rounds. That's good. Nevertheless there is another issue. In my example above with the two hashers, [this code](https://github.com/mohiva/play-silhouette/blob/22dfddecaccae8139f80e7a19c571ee8fc299157/silhouette/app/com/mohiva/play/silhouette/impl/providers/CredentialsProvider.scala#L69-L73) will do the wrong thing. In particular `hasher != passwordHasher` will be false and so old hashes will not be updated to become stronger.

What test practices are you worried about violating? It's not like you're calling an external service, and hashing with ~1000 rounds won't be slow enough to notice.

Is it possible to include the BCrypt package as a 'test dependency' that doesn't interfere with production builds? I don't know much about sbt.

akkie:
Indeed this is an issue. We could override the `equals` method and use the rounds as an equals factor.

> What test practices are you worried about violating?

We would use a concrete implementation of a dependency for a unit test. What if we have 10 other password hasher implementations? Would you also create tests for these implementations? If we must do that then this a sign for a wrong software architecture. The `CredentialsProvider` doesn't know the existence of a `BCryptPasswordHasher` why should the test know the existence of this class?

If we override the `equals` method then we can test this case also with a mock and there is no need to bind the concrete implementation to a test.

alexmojaki:
You should only need to put logic that identifies and distinguishes different hashers into one method: either `id` or `equals`. I think `id` is much simpler.

Tests including concrete implementations of hashers would really be testing the correctness of the hashers, not `CredentialsProvider`. I think it should be documented in `PasswordHasher` that `id` must be implemented so that if two hashers don't match then they must return different values for `id`. Currently `BCryptPasswordHasher` fails to fulfil this contract, and a test dealing with that makes sense to me.

However you are right that this would get repetitive if there were many implementations which indicate that there should be a better way. I have committed a way which could work: ensure on construction of the `CredentialsProvider` that all the hashers have unique IDs. That would throw an exception on startup in my initial example. In all cases the exception would indicate that either the provider was constructed incorrectly or the hashers were implemented incorrectly. This would also help people who write their own implementations of `PasswordHasher`.

akkie:
Please see #456

PR #456, opened 2016-03-30
Password hashers should not only be equal based on types

akkie:
Fixes https://github.com/mohiva/play-silhouette/pull/454

### DIFF
--- a/silhouette-password-bcrypt/src/main/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasher.scala
+++ b/silhouette-password-bcrypt/src/main/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasher.scala
@@ -26,10 +26,14 @@ import org.mindrot.jbcrypt.BCrypt
 /**
  * Implementation of the password hasher based on BCrypt.
  *
+ * We declare this class as case class so that the compiler can create the `equals` method for us. This
+ * is needed for the credentials based providers as an indicator that a password should be re-hashed,
+ * if the log rounds value was changed.
+ *
  * @param logRounds The log2 of the number of rounds of hashing to apply.
  * @see [[http://www.mindrot.org/files/jBCrypt/jBCrypt-0.2-doc/BCrypt.html#gensalt(int) gensalt]]
  */
-class BCryptPasswordHasher(logRounds: Int = 10) extends PasswordHasher {
+case class BCryptPasswordHasher(logRounds: Int = 10) extends PasswordHasher {
 
   /**
    * Gets the ID of the hasher.
--- a/silhouette-password-bcrypt/src/test/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasherSpec.scala
+++ b/silhouette-password-bcrypt/src/test/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasherSpec.scala
@@ -53,4 +53,20 @@ class BCryptPasswordHasherSpec extends Specification {
       hasher.matches(info, "not-equal") must beFalse
     }
   }
+
+  "The hasher" should {
+    "be equal if the rounds parameter is equal" in {
+      val hasher1 = new BCryptPasswordHasher(1)
+      val hasher2 = new BCryptPasswordHasher(1)
+
+      hasher1 == hasher2 must beTrue
+    }
+
+    "not be equal if the rounds parameter is not equal" in {
+      val hasher1 = new BCryptPasswordHasher(1)
+      val hasher2 = new BCryptPasswordHasher(2)
+
+      hasher1 == hasher2 must beFalse
+    }
+  }
 }
--- a/silhouette/app/com/mohiva/play/silhouette/api/util/PasswordHasher.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/api/util/PasswordHasher.scala
@@ -32,6 +32,13 @@ case class PasswordInfo(hasher: String, password: String, salt: Option[String] =
 
 /**
  * A trait that defines the password hasher interface.
+ *
+ * Some credentials based providers have the ability to change the password hashing method on
+ * the fly. These providers use the `equals` method of a hasher as an indicator that the password
+ * should be re-hashed with a new hashing algorithm. It can be the case that an hasher can provide
+ * higher security based on additional parameters. In this case a default equals check will not work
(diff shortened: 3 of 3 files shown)


alexmojaki:
When I said

> In particular `hasher != passwordHasher` will be false

I may have misdirected your attention. The reason the code fails is not that the `equals` method isn't implemented. I'm actually pretty sure that the default implementation will do just fine. The problem is that [this line](https://github.com/mohiva/play-silhouette/blob/22dfddecaccae8139f80e7a19c571ee8fc299157/silhouette/app/com/mohiva/play/silhouette/impl/providers/CredentialsProvider.scala#L69) will (sometimes) find the wrong hasher, hence my focus on `id`, and ultimately that will mess with the if statement.

I'd like to reiterate that this would be apparent if there was a test involving both `CredentialsProvider` and `BCryptPasswordHasher`.

akkie:
> The reason the code fails is not that the equals method isn't implemented. I'm actually pretty sure that the default implementation will do just fine.

```scala
class Hasher(round: Int)

val hasher1 = new Hasher(1)
val hasher2 = new Hasher(1)

hasher1
hasher2
hasher1 == hasher2

case class EqualsHasher(round: Int)

val equalsHasher1 = EqualsHasher(1)
val equalsHasher2 = EqualsHasher(1)

equalsHasher1
equalsHasher2
equalsHasher1 == equalsHasher2
```

```
defined class Hasher

hasher1: Hasher = Hasher@48bd99af
hasher2: Hasher = Hasher@61874688

res0: Hasher = Hasher@48bd99af
res1: Hasher = Hasher@61874688
res2: Boolean = false

defined class EqualsHasher

equalsHasher1: EqualsHasher = EqualsHasher(1)
equalsHasher2: EqualsHasher = EqualsHasher(1)

res3: EqualsHasher = EqualsHasher(1)
res4: EqualsHasher = EqualsHasher(1)
res5: Boolean = true
```
Sometimes code says more than thousand words.

> The problem is that this line will (sometimes) find the wrong hasher, hence my focus on id, and ultimately that will mess with the if statement.

I see your point! But the solution with the ID should be well considered, because it would break backward compatibility and can lead to a lot of headache on the user side.

Consider the following scenario:
Every credentials based auth info is stored with the `bcrypt` ID. If we update the ID to include also the rounds, then we invalidate all password entries in the DB.

With your solution, the user must also manage every used BCrypt hasher in the code, also when it's not really needed, because every BCrypt password hasher can match every BCrypt hashed password, regardless of what rounds values was used.

So an instantiation of the `CredentialsProvider` with the following values would always work.

```
new CredentialsProvider(..., new BCryptPasswordHasher(20), Seq(new BCryptPasswordHasher()))
```

I know this may only be true for the `BCryptPasswordHasher`. But I will only show you that a solution is not so easy as it seems. And it shows you also that it brings nothing to test the `CredentialsProvider` with the `BCryptPasswordHasher` because the test would only be true for this type of hasher. So we need a solution which works with all kind of password hasher implementations.

> I'd like to reiterate that this would be apparent if there was a test involving both CredentialsProvider and BCryptPasswordHasher.

In the `CredentialsProviderSpec` you can  construct every possible combination with the help of different `PasswordHasher` mocks. So there is no need to use concrete implementations like the `BCryptPasswordHasher`.

alexmojaki:
I am well aware that the default `equals` is equivalent to `eq`. I am asserting that in the code samples I've seen, this is good enough. However you have made me realise that someone might instantiate equivalent hashers twice which would lead to unnecessary updates, so something along these lines is still needed to keep the code robust and efficient. However it is still essential to distinguish hashers based on `id`. Earlier I said that such logic needs to be in either `equals` or `id`, and that was wrong: `id` must be fully implemented. There is no other way that the provider can know that the hashes must be updated. Without proper IDs, people will think that they are upgrading the security of their databases, but nothing will happen.

I understand that this will break compatibility, but I think this is an acceptable sacrifice. The major version is increasing and everyone expects things to break if they are upgrading. There will be an exception with the nice `UnsupportedHasher` message. The source code here is easy to read so it wouldn't be hard to infer the problem. Nevertheless there can definitely be a mention in the changelog. In fact now that the hasher is in a separate package, code using it won't even compile and users will almost certainly hunt down documentation to know what needs changing. If this issue is mentioned in close proximity I doubt there will be headaches.

For users to upgrade smoothly, there can be a legacy hasher which has the simple `"bcrypt"` ID and delegates to an actual hasher. Such a utility class can be provided here. Alternatively it should be quite simple to update a database and replace all instances of `bcrypt` with `bcrpyt10` or whatever. It's a bit of work but it's not a disaster in exchange for the code doing the right thing.

Testing with mocks can only get you so far. In particular it only works if the concrete implementations actually have the same properties as the mocks. `CredentialsProvider` works fine, it doesn't need to be changed or tested further, and adding more tests with mocks wouldn't add any useful information. We already know that it does the right thing when hashers have different IDs thanks to an existing test using mocks. That's great. What needs to be tested is that it does the right thing with `BCryptPasswordHasher` in particular to show that the hasher is correctly implemented. If you had such a test earlier you would have been able to see that this PR didn't solve anything. I can't imagine what test with mocks would have accomplished that. There needs to be some black-box testing to show that all the pieces are fitting together correctly.

akkie:
After thinking about the problem I came to the conclusion, that no further changes are needed. There is currently not really a problem with the `CredentialsProvider` as well with the `BCryptPasswordHasher`.

As noted in my last post the `BCryptPasswordHasher` can match all password hashes regardless of what round values was used and I have show you a way how you could solve your initial issue.

Other password hasher implementations can use the same strategy as the `BCryptPasswordHasher` if the algorithm itself supports the same behavior. If not they can use a strategy based on the provider ID.

All this is currently possible with the current `CredentialsProvider` implementation and I see no need to break backward compatibility for a not existent issue.

If you find an issue I currently not see, then please open a new issue and provide a failing test case.

Currently I see no need to integrate integration tests into this repository, then this is what you propose. It's also not possible with the library design to test the `CredentialsProvider` with the `BCryptPasswordHasher` because then we had a circular dependency between the `play-silhouette` and the `play-silhouette-password-bcrypt` project. It would be possible to implement integration test into the seed template. So it would be nice if you can create a pull request with such a test.

If you find a simple way to integrate integration tests so that it makes sense for the complete library and not only for this particular issue then a pull request would also be welcome.

PR #459, opened 2016-04-04
Added failing test to show that BCryptPasswordHasher needs distinct IDs

alexmojaki:
Related: #456 and #454 

Adding `logRounds` to the `id` fixes the test.

The test will pass if the hasher list is reversed, i.e. `List(oldHasher, newHasher)`, meaning that this bug is not 'well-behaved'.

The test still fails if the hasher list is simply `List(newHasher)`.

Making the hasher a case class makes no difference.

I expect that the code for the test can be improved as this is my first time dealing with these libraries.

### DIFF
--- a/silhouette-password-bcrypt/build.sbt
+++ b/silhouette-password-bcrypt/build.sbt
@@ -2,7 +2,8 @@ import Dependencies._
 
 libraryDependencies ++= Seq(
   Library.jbcrypt,
-  Library.Specs2.core % Test
+  Library.Specs2.core % Test,
+  Library.Play.Specs2.mock % Test
 )
 
 enablePlugins(Doc)
--- a/silhouette-password-bcrypt/src/main/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasher.scala
+++ b/silhouette-password-bcrypt/src/main/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasher.scala
@@ -36,7 +36,7 @@ class BCryptPasswordHasher(logRounds: Int = 10) extends PasswordHasher {
    *
    * @return The ID of the hasher.
    */
-  override def id = ID
+  override def id = ID + logRounds
 
   /**
    * Hashes a password.
--- a/silhouette-password-bcrypt/src/test/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasherSpec.scala
+++ b/silhouette-password-bcrypt/src/test/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasherSpec.scala
@@ -35,11 +35,12 @@ class BCryptPasswordHasherSpec extends Specification with Mockito {
   "The `hash` method" should {
     "hash a password" in {
       val password = "my_S3cr3t_p@sswQrd"
-      val hasher = new BCryptPasswordHasher
+      val logRounds = 10
+      val hasher = new BCryptPasswordHasher(logRounds)
       val info = hasher.hash(password)
 
       info must beAnInstanceOf[PasswordInfo]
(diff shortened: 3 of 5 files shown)


alexmojaki:
This conflicts with commits from #456 because github did funny things with my forks but my understanding is that those changes will be removed anyway.

akkie:
The following test works as expected for me:
```scala
val credentials = Credentials("apollonia.vanova@watchmen.com", "s3cr3t")
val authInfoRepository = mock[AuthInfoRepository]
val oldHasher = new BCryptPasswordHasher(5)
val newHasher = new BCryptPasswordHasher(7)
val provider = new CredentialsProvider(authInfoRepository, newHasher, List(oldHasher))
val loginInfo = LoginInfo(provider.id, credentials.identifier)
val passwordInfo = oldHasher.hash(credentials.password)

authInfoRepository.find[PasswordInfo](loginInfo) returns Future.successful(Some(passwordInfo))
authInfoRepository.update[PasswordInfo](any, any) answers { (a, m) =>
  Future.successful(a.asInstanceOf[Array[Any]](1).asInstanceOf[PasswordInfo])
}

Await.result(provider.authenticate(credentials), 5.seconds) must be equalTo loginInfo

there was one(authInfoRepository).update(any, any)
```
I think your test fails with `List(newHasher)` because your repository misses https://github.com/mohiva/play-silhouette/pull/456

alexmojaki:
The important difference in your test is here:

`new CredentialsProvider(authInfoRepository, newHasher, List(oldHasher))`

which is certainly a strange way to interact with the API. I would say it only works by accident. Is that what you intended to write?

akkie:
> which is definitely bad because it will always trigger an update.

This is indeed an issue I hadn't thought off. But let us consider if we find a solution which doesn't break backward compatibility. If not, then we could use the ID approach.

alexmojaki:
OK, I've added a solution that should preserve backwards compatibility by fixing `PasswordInfo`s as they are retrieved from a database. It's not perfect, but it's an idea.

akkie:
@alexmojaki What do you think about something like [this](https://github.com/mohiva/play-silhouette/pull/461). It defines a better interface for this functionality. It would break binary compatibility but the user mustn't update the database. It allows also better flexibility for other password hashers.

If you increase the log rounds of the BCrypt hasher then you must only pass the current hasher with the increased value. The hasher recognizes that the password info is deprecated and the provider hashes the password again. There is no need to pass a list of BCrypt hashers like:
```
Seq(BCryptPasswordHasher(10), BCryptPasswordHasher(12), BCryptPasswordHasher(16), ...)
```

alexmojaki:
I'll comment in more detail later, but my immediate thoughts are that this introduces far too much complexity to a simple area. This library is already a lot to take in, please try and keep the API lightweight and easy to learn. Also, the user does have to update the database because they have to add a column for metadata.

akkie:
I've added a description with some examples, which should show that the new API is much simpler to understand for the end-user.

> I'll comment in more detail later, but my immediate thoughts are that this introduces far too much complexity to a simple area

Could you please describe the complex points you see? And I think it has showed us that this is not really a simple area. It has many pitfalls and I think the new API can handle these in an easy and understandable manner. It's now also easier to test. And this is what I meant as I said that the architecture of the library is not well designed if it's needed to use a concrete dependency to test a class. Because in this case the dependency contains logic which has impact of the class to test. Instead only the interface should have impact of the class to test. I'm sure that the new implementation doesn't violate against a single SOLID principle.

> Also, the user does have to update the database because they have to add a column for metadata.

Yes the column must be added if a relational SQL database is used. But the user mustn't update the complete data set.

akkie:
See #461

PR #461, opened 2016-04-06
Define meaningful interface for password re-hashing

akkie:
WIP

In contrast to the current implementation it now defines the clear responsibility for every component. Currently this is defined squishy in the interface and therefore can lead to bugs like #454 or #459.

What are the changes for the user:
It's now unambiguous when a password hasher or a password is deprecated and when it should be updated with a newer algorithm. This is now defined through the `PasswordHasherRegistry`. It accepts two parameters. The current preferred algorithm and the list of deprecated hashers which are needed to macht the old hashes which are stored in the database.

So to instantiate the `CredentialsProvider` as example, a user must now write:
```scala
new CredentialsProvider(..., new PasswordHasherRegistry(new BCryptPassordHasher(10)))
```

instead of

```scala
new CredentialsProvider(..., new BCryptPassordHasher(10), Seq(new BCryptPassordHasher(10)))
```

I can now handle the case of a deprecated password for a not deprecated password hasher as described in #454.

The user can only increase the log rounds for a password hasher:
```scala
new CredentialsProvider(..., new PasswordHasherRegistry(new BCryptPassordHasher(20)))
```

If another password hashing algorihtm should be used, then it's now clear defined what hasher is preferred and which is deprecated.
```scala
new CredentialsProvider(..., new PasswordHasherRegistry(new SCryptPassordHasher(), Seq(new BCryptPassordHasher())))
```

instead of

```scala
new CredentialsProvider(..., new SCryptPassordHasher(), Seq(new SCryptPassordHasher(), new BCryptPassordHasher())))
```

This pull request introduces a new optional `metadata` field for the `PasswordInfo` class. so if you need a relational SQL database, then you must add this field to your database table. This additional field is needed to store additional data for a hasher which can be used to check if a password is deprecated or not. It's not needed to convert you data in the database.

### DIFF
--- a/silhouette-password-bcrypt/src/main/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasher.scala
+++ b/silhouette-password-bcrypt/src/main/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasher.scala
@@ -26,14 +26,10 @@ import org.mindrot.jbcrypt.BCrypt
 /**
  * Implementation of the password hasher based on BCrypt.
  *
- * We declare this class as case class so that the compiler can create the `equals` method for us. This
- * is needed for the credentials based providers as an indicator that a password should be re-hashed,
- * if the log rounds value was changed.
- *
  * @param logRounds The log2 of the number of rounds of hashing to apply.
  * @see [[http://www.mindrot.org/files/jBCrypt/jBCrypt-0.2-doc/BCrypt.html#gensalt(int) gensalt]]
  */
-case class BCryptPasswordHasher(logRounds: Int = 10) extends PasswordHasher {
+class BCryptPasswordHasher(logRounds: Int = 10) extends PasswordHasher {
 
   /**
    * Gets the ID of the hasher.
@@ -51,7 +47,10 @@ case class BCryptPasswordHasher(logRounds: Int = 10) extends PasswordHasher {
    * @param plainPassword The password to hash.
    * @return A PasswordInfo containing the hashed password.
    */
-  override def hash(plainPassword: String) = PasswordInfo(id, BCrypt.hashpw(plainPassword, BCrypt.gensalt(logRounds)))
+  override def hash(plainPassword: String) = PasswordInfo(
+    hasher = id,
+    password = BCrypt.hashpw(plainPassword, BCrypt.gensalt(logRounds))
+  )
 
   /**
    * Checks if a password matches the hashed version.
@@ -62,6 +61,24 @@ case class BCryptPasswordHasher(logRounds: Int = 10) extends PasswordHasher {
    */
   override def matches(passwordInfo: PasswordInfo, suppliedPassword: String) = {
     BCrypt.checkpw(suppliedPassword, passwordInfo.password)
+  }
+
+  /**
+   * Indicates if a password info hashed with this hasher is deprecated.
+   *
+   * In case of the BCrypt password hasher, a password is deprecated if the log rounds have changed.
+   *
+   * @param passwordInfo The password info to check the deprecation status for.
+   * @return True if the given password info is deprecated, false otherwise. If a hasher isn't
+   *         suitable for the given password, this method should return None.
+   */
+  override def isDeprecated(passwordInfo: PasswordInfo): Option[Boolean] = {
+    Option(isSuitable(passwordInfo)).collect {
+      case true =>
+        val LogRoundsPattern(lr) = passwordInfo.password
+        // Is deprecated if the log rounds has changed
+        lr != logRounds.toString
+    }
   }
 }
 
@@ -74,4 +91,9 @@ object BCryptPasswordHasher {
    * The ID of the hasher.
    */
   val ID = "bcrypt"
+
+  /**
+   * The pattern to extract the log rounds from the password string.
+   */
+  val LogRoundsPattern = """^\$\w{2}\$(\d{1,2})\$.+""".r
 }
--- a/silhouette-password-bcrypt/src/test/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasherSpec.scala
+++ b/silhouette-password-bcrypt/src/test/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasherSpec.scala
@@ -17,6 +17,7 @@ package com.mohiva.play.silhouette.password
 
 import com.mohiva.play.silhouette.api.util.PasswordInfo
 import org.specs2.mutable.Specification
+import org.specs2.specification.Scope
 
 /**
  * Test case for the [[BCryptPasswordHasher]] class.
@@ -24,49 +25,94 @@ import org.specs2.mutable.Specification
 class BCryptPasswordHasherSpec extends Specification {
 
   "The `hash` method" should {
-    "hash a password" in {
-      val password = "my_S3cr3t_p@sswQrd"
-      val hasher = new BCryptPasswordHasher
-      val info = hasher.hash(password)
+    "hash a password" in new Context {
+      val passwordInfo: PasswordInfo = hasher.hash(password)
 
-      info must beAnInstanceOf[PasswordInfo]
-      info.hasher must be equalTo BCryptPasswordHasher.ID
-      info.password must not be equalTo(password)
-      info.salt must beNone
+      passwordInfo.hasher must be equalTo BCryptPasswordHasher.ID
+      passwordInfo.password must not be equalTo(password)
+      passwordInfo.salt must beNone
     }
   }
 
   "The `matches` method" should {
-    "return true if a password matches a previous hashed password" in {
-      val password = "my_S3cr3t_p@sswQrd"
-      val hasher = new BCryptPasswordHasher
-      val info = hasher.hash(password)
+    "return true if a password matches a previous hashed password" in new Context {
+      val passwordInfo = hasher.hash(password)
 
-      hasher.matches(info, password) must beTrue
+      hasher.matches(passwordInfo, password) must beTrue
     }
 
-    "return false if a password doesn't match a previous hashed password" in {
-      val password = "my_S3cr3t_p@sswQrd"
-      val hasher = new BCryptPasswordHasher
-      val info = hasher.hash(password)
+    "return true if a password matches a previous hardcoded password" in new Context {
+      val passwordInfo = PasswordInfo("bcrypt", "$2a$10$PBwXy.iQz9n4QOdbgEV7Ve2aYsvXeAvyT0rzhoZKwaDH/3j3tUSW.")
 
-      hasher.matches(info, "not-equal") must beFalse
+      hasher.matches(passwordInfo, password) must beTrue
+    }
+
+    "return false if a password doesn't match a previous hashed password" in new Context {
+      val passwordInfo = hasher.hash(password)
+
+      hasher.matches(passwordInfo, "not-equal") must beFalse
     }
   }
 
-  "The hasher" should {
-    "be equal if the rounds parameter is equal" in {
-      val hasher1 = new BCryptPasswordHasher(1)
-      val hasher2 = new BCryptPasswordHasher(1)
+  "The `isSuitable` method" should {
+    "return true if the hasher is suitable for the given password info" in new Context {
+      val passwordInfo = hasher.hash(password)
 
-      hasher1 == hasher2 must beTrue
+      hasher.isSuitable(passwordInfo) must beTrue
     }
 
-    "not be equal if the rounds parameter is not equal" in {
-      val hasher1 = new BCryptPasswordHasher(1)
-      val hasher2 = new BCryptPasswordHasher(2)
+    "return true if the hasher is suitable when given a password info with different log rounds" in new Context {
+      val currentHasher = new BCryptPasswordHasher(5)
 
-      hasher1 == hasher2 must beFalse
+      val storedHasher = new BCryptPasswordHasher(10)
+      val passwordInfo = storedHasher.hash(password)
+
+      currentHasher.isSuitable(passwordInfo) must beTrue
     }
+
+    "return false if the hasher isn't suitable for the given password info" in new Context {
+      val passwordInfo = PasswordInfo("scrypt", "")
+
+      hasher.isSuitable(passwordInfo) must beFalse
+    }
+  }
+
+  "The `isDeprecated` method" should {
+    "return None if the hasher isn't suitable for the given password info" in new Context {
+      val passwordInfo = PasswordInfo("scrypt", "")
+
+      hasher.isDeprecated(passwordInfo) must beNone
+    }
+
+    "return Some(true) if the stored log rounds are not equal the hasher log rounds" in new Context {
+      val currentHasher = new BCryptPasswordHasher(5)
+
+      val storedHasher = new BCryptPasswordHasher(10)
+      val passwordInfo = storedHasher.hash(password)
+
+      currentHasher.isDeprecated(passwordInfo) must beSome(true)
+    }
+
+    "return Some(false) if the stored log rounds are equal the hasher log rounds" in new Context {
+      val passwordInfo = hasher.hash(password)
+
+      hasher.isDeprecated(passwordInfo) must beSome(false)
+    }
+  }
+
+  /**
+   * The context.
+   */
+  trait Context extends Scope {
+
+    /**
+     * A plain text password.
+     */
+    lazy val password = "my_S3cr3t_p@sswQrd"
+
+    /**
+     * The hasher to test.
+     */
+    lazy val hasher = new BCryptPasswordHasher(10)
   }
 }
--- a/silhouette/app/com/mohiva/play/silhouette/api/util/PasswordHasher.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/api/util/PasswordHasher.scala
@@ -28,17 +28,14 @@ import com.mohiva.play.silhouette.api.AuthInfo
  * @param password The hashed password.
  * @param salt The optional salt used when hashing.
  */
-case class PasswordInfo(hasher: String, password: String, salt: Option[String] = None) extends AuthInfo
+case class PasswordInfo(
+  hasher: String,
+  password: String,
+  salt: Option[String] = None
+) extends AuthInfo
 
 /**
  * A trait that defines the password hasher interface.
- *
- * Some credentials based providers have the ability to change the password hashing method on
- * the fly. These providers use the `equals` method of a hasher as an indicator that the password
- * should be re-hashed with a new hashing algorithm. It can be the case that an hasher can provide
- * higher security based on additional parameters. In this case a default equals check will not work
- * because the hasher instance is the same. In such a case you should override the equals method
- * to provide a better check based on the additional parameters.
  */
 trait PasswordHasher {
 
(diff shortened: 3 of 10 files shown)


alexmojaki:
OK, I think this is good, but it can be simplified both to ease migration for upgrading users and to keep things easy for new users.

First, the metadata field isn't needed. Such data can still be stored in the ID, and the new hasher methods can be implemented by parsing that. There is no need to separate the algorithm used and the parameters of the algorithm in the database. Now users don't have to update the database.

There is also no need for the `PasswordHasherRegistry`. It just adds more things that both users and developers need to know about. The `CredentialsProvider` is perfectly capable of handling that logic itself and taking two arguments as before. The only change is that the user is no longer required to include the current hasher in the hasher list, although if they do, it should still be handled properly. Then users don't have to rewrite any code.

akkie:
Thanks for your help and your inspiring suggestion. But I've decided to keep the code as it is.

For me, to encode the data in the ID is a hack. Also the usage of the metadata field allows further hashers to store other data. Maybe a hasher has the need to store more than a single value. So it can store a serialized object in this field.

The `PasswordHasherRegistry` is not only needed by the `CredentialsProvider`. The `BasicAuthProvider` uses the same logic. So it encapsulates logic that is needed by both providers.

alexmojaki:
> For me, to encode the data in the ID is a hack.

How? The only problem to me is that the name sounds a bit off, and not even that much. The point is  that you only need one field to *identify* the hasher, to define how it works, and to distinguish it from others.

> Also the usage of the metadata field allows further hashers to store other data. Maybe a hasher has the need to store more than a single value. So it can store a serialized object in this field.

So? How is using two string fields easier than one? Is storing 4 parameters in one string field and 1 parameter in a second easier than just storing 5 parameters in one field?

Ultimately there is no actual important distinction between the algorithm used and the parameters of the algorithm. As humans we feel like they look different, but in terms of writing code they're the same. They're all just parameters of the larger question "how do I hash this password". If I had needed to implement my own hasher for a different algorithm I would have encoded the parameters in the ID without a second thought. It seems perfectly natural to me.

If you really feel that there should be two separate parameters, make sure that you have a good reason, not just that it feels odd, so that no one needs to write a database migration and new ORM mappings and so on.

> The `PasswordHasherRegistry` ... encapsulates logic that is needed by both providers.

OK, that's a good reason, I hadn't realised. However, that's an implementation detail, not something that users need to see. The correct approach would be to leave the API as is, let the providers construct the registry themselves early on, and make `PasswordHasherRegistry` a package-private class.

However, you can do much better. The two providers are extremely similar and much more of them needs to be factored out into a class that they both make use of. The registry is barely doing anything to help deal with the duplication between the classes.

akkie:
For me the ID is an identifier of the hasher. All the bcrypt hashers should have the same ID. This is my point of view and you have another. It's also the same as for all other providers in Silhouette. I think we could discuss this forever and the meaning wouldn't change. So I think we should ask for another meaning.

alexmojaki:
> For me the ID is an identifier of the hasher

Yes. Is a `BCryptPasswordHasher(5)` the same as a `BCrpytPasswordHasher(10)`? No, so they should be identified differently.

If you don't like thinking of it that way, then don't worry about it. The point is that you just need one field from which the code can obtain the correct hasher to work with. It's as if the field contains a serialised form of the hasher.

The ID in its current form is weird anyway. Why create a constant for each implementation of `PasswordHasher`? The class name would have sufficed.

In any case these arguments about semantics are pointless. Focus on the practical implications. You have an opportunity to make this perfectly backward compatible. Is your distinction between algorithm and parameters important enough to sacrifice that?

If you truly insist on having a metadata field, at least ensure that it introduces a compile time error (perhaps by removing the default). Right now it's easy for people to not be aware that they need to update their database and other code. Depending on their code, they'll either get a runtime error when it doesn't know how to store the metadata, or the extra field will be ignored and the issue will continue.

akkie:
A few days are gone now, and I still think that the metadata field is the better solution. Decoding the parameters in the ID is to limited. What if a hasher has the need to store multiple parameters and one of those parameters can be an arbitrary string value. What if the user passes a string like 'my-hasher'? Or some other string which contains the same spacer which is used to separate the parameters in the ID. Sure you could implement an escaping strategy, but this could also be error prone.

The reason why we use a constant instead of the class name is that as first, a class name can change and as second, that the database can also be used from another language which has a different naming scheme. So the string value we define for the hasher is a good compromise I think.

I think it's a good idea to remove the default value for the metadata field so that a user gets a compile time error.

alexmojaki:
> What if a hasher has the need to store multiple parameters and one of those parameters can be an arbitrary string value.

We dealt with this question a while ago.

>> Also the usage of the metadata field allows further hashers to store other data. Maybe a hasher has the need to store more than a single value. So it can store a serialized object in this field.

> So? How is using two string fields easier than one? Is storing 4 parameters in one string field and 1 parameter in a second easier than just storing 5 parameters in one field?

There are plenty of ways to store arbitrary data in a string safely and easily. JSON, for example, is more than sufficient.

Those are indeed good reasons to have a string constant per hasher.

akkie:
> We dealt with this question a while ago.

Have we really talked about arbitrary data in the ID? If so, then sorry!

> There are plenty of ways to store arbitrary data in a string safely and easily. JSON, for example, is more than sufficient.

Yes this is possible, but then we must convert the existing IDs to follow this scheme. Or would you store different formats for different hashers? I think an ID should follow the same scheme.

alexmojaki:
> Have we really talked about arbitrary data in the ID?

Yes, I just quoted it. Here is us talking about it:

>> Also the usage of the metadata field allows further hashers to store other data. Maybe a hasher has the need to store more than a single value. So it can store a serialized object in this field.

> So? How is using two string fields easier than one? Is storing 4 parameters in one string field and 1 parameter in a second easier than just storing 5 parameters in one field?

-------

Back to the present:

I think that the `id` method should be made concrete and final (to force existing hashers to migrate) to return the following JSON: `{ "name" : $name(), "params" : $params() }`. `name` is the equivalent of the current `id`, generally returning a constant such as `"bcrypt"`, while `params` is like `metadata` (or you can keep the name `metadata` if you want, but it doesn't strike me as particularly meta). `params` can return a string, a number, None, or anything JSON serializable.

akkie:
But then a user must update the complete data set to follow this scheme. I really see no advantage over the current implementation. Only disadvantages:
- The user must update the complete data set
- The user cannot _easily_ query the list of hashers (A user must search for a string in a string, no index)
- It would make further updates more difficult(Same as above)

I'll go with the current implemented solution. Do you have any notes to the source code? Otherwise I would merge the pull request.

alexmojaki:
> - The user must update the complete data set

The code could probably be written to recognise whether or not the ID coming from the database is following the new scheme and behave accordingly. If you're worried that this is too complex, then let each hasher simply follow its own scheme for encoding name and params in the ID. I think that the chance of two hashers conflicting (i.e. both returning `true` from `isSuitable` when only one is appropriate) is very small unless the implementation for the hasher is really useless. It's sufficient for a hasher to ensure that the ID starts with a unique string (the name).

In any case, your current implementation requires that the user alters the schema of the database.

> - The user cannot easily query the list of hashers (A user must search for a string in a string, no index)
> - It would make further updates more difficult(Same as above)

I can't really see a likely scenario where this would be a significant problem. Why would the user want to find rows that have been hashed in a particular way? Even if they did, a full table scan is not a significant problem. The user is not looking for a single row, they probably want to return at least 10% of the table.

> I'll go with the current implemented solution. Do you have any notes to the source code? Otherwise I would merge the pull request.

It feels like all previous comments in this discussion are being discarded. Yesterday you agreed:

> I think it's a good idea to remove the default value for the metadata field so that a user gets a compile time error.

And moreover, you're not addressing the fact that the user is forced to do a schema migration. Are you really that comfortable with the fact?

akkie:
Yes, I'm comfortable with the schema migration. From your comments I see that you try to workaround this issue, but it remains a workaround. And currently we jump from one workaround to another.

The current implementation solves this issue in a clean manner. The only disadvantage I see is the schema migration. But I think it's better to force the user to migrate the schema as to implement a dirty solution. It's a major release and the schema migration should be a very easy task.

> It feels like all previous comments in this discussion are being discarded. Yesterday you agreed:

>> I think it's a good idea to remove the default value for the metadata field so that a user gets a compile time error.
> And moreover, you're not addressing the fact that the user is forced to do a schema migration. Are you really that comfortable with the fact?

I've updated the pull request yesterday and removed the default value. If I've merged the pull request, I will also adapt the migration guide to point to the schema migration.

To avoid endless discussions, I do not react to all your comments, because most of the time we have a completely different meaning of software design. This isn't against you in personal. It's only to save my time. As you surely noted, English isn't my native langue and it costs my a lot of time to write these conversations with you. I hope you can understand this.

silhouette-password-bcrypt/src/main/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasher.scala
```diff
@@ -51,7 +47,11 @@ case class BCryptPasswordHasher(logRounds: Int = 10) extends PasswordHasher {
    * @param plainPassword The password to hash.
    * @return A PasswordInfo containing the hashed password.
    */
-  override def hash(plainPassword: String) = PasswordInfo(id, BCrypt.hashpw(plainPassword, BCrypt.gensalt(logRounds)))
+  override def hash(plainPassword: String) = PasswordInfo(
+    hasher = id,
+    password = BCrypt.hashpw(plainPassword, BCrypt.gensalt(logRounds)),
+    metadata = Some(logRounds.toString)
```
alexmojaki: The metadata is actually more complicated because it should also include the version of the bcrypt algorithm. bcrypt hashes follow a [modular crypt format](https://pythonhosted.org/passlib/modular_crypt_format.html):

> In short, the modular crypt format (MCF) is a standard for encoding password hash strings, which requires hashes have the format `$identifier$content`; where `identifier` is an short alphanumeric string uniquely identifying a particular scheme, and `content` is the contents of the scheme, using only the characters in the regexp range `[a-zA-Z0-9./]`.

Currently `identifier = "2a$" + logRounds`. The `2a` part can be changed in the future if the algorithm is updated, so that should be stored. I think a decent future-proof method would be `metadata = password.substring(1, password.lastIndexOf('$'))`.

Note the redundancy here: metadata is computed from the password field! This is why I think it's odd that you want to add a column. `BCryptPasswordHasher` is perfectly capable of implementing `isDeprecated` and `isSuitable` based on `PasswordInfo` as it is now without `metadata` and without even needing to change `id`. Really it would take quite an awful hasher implementation to be able to implement `matches` correctly but not `isSuitable` and `isDeprecated` based only on the current `PasswordInfo`. Between `id` and `password` I think there is plenty of space to store as much metadata as needed without causing any inconvenience. Currently `password` is comfortably storing 4 pieces of data in a single string for bcrypt.

alexmojaki:
Sorry, I hadn't noticed you updated the pull request because there was still only one commit. Did you rewrite the history?

Anyway, I understand that it's a pain for you to continue this discussion. You don't have to. I put a comment in the code about metadata but it's OK if you don't respond to the rant at the end.

I haven't had a chance to review all the code properly. I'll let you know if there's anything else wrong or if you can go ahead.

silhouette-password-bcrypt/src/main/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasher.scala
akkie: You are completely right! This is my [first comment of your initial issue](https://github.com/mohiva/play-silhouette/pull/454#issuecomment-201814287):

> A bcrypt password hash includes the rounds the hash wars created with. So there is no need to save the number of rounds otherwise.

I've absolutely no idea why I came to the idea to store the log rounds in an extra field. Maybe because of #459.

I'll update the pull request so that the `isDeprecated` method extracts the log rounds from the password hash.

akkie:
> Did you rewrite the history?

Yes I have squashed the commits with `git rebase`.

akkie:
I've updated the pull request.

silhouette-password-bcrypt/src/test/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasherSpec.scala
```diff
@@ -17,56 +17,82 @@ package com.mohiva.play.silhouette.password
...
-      val password = "my_S3cr3t_p@sswQrd"
-      val hasher = new BCryptPasswordHasher
-      val info = hasher.hash(password)
-
-      info must beAnInstanceOf[PasswordInfo]
-      info.hasher must be equalTo BCryptPasswordHasher.ID
-      info.password must not be equalTo(password)
-      info.salt must beNone
+    "hash a password" in new Context {
+      override val passwordInfo = hasher.hash(password)
```
alexmojaki: The type is statically determined, just declare it here rather than `beAnInstanceOf`.

alexmojaki: Why is it being overridden? What's wrong with the one in the context?

silhouette-password-bcrypt/src/test/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasherSpec.scala
```diff
@@ -17,56 +17,82 @@ package com.mohiva.play.silhouette.password
...
+    "return true if a password matches a previous hashed password" in new Context {
+      hasher.matches(passwordInfo, password) must beTrue
+    }
+
+    "return false if a password doesn't match a previous hashed password" in new Context {
+      hasher.matches(passwordInfo, "not-equal") must beFalse
+    }
+  }
 
-      hasher.matches(info, password) must beTrue
```
alexmojaki: I think a test of this nature (i.e. that `matches` and `hash` are compatible) is still useful.

silhouette-password-bcrypt/src/test/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasherSpec.scala
```diff
@@ -17,56 +17,82 @@ package com.mohiva.play.silhouette.password
...
+  "The `isSuitable` method" should {
+    "return true if the hasher is suitable for the given password info" in new Context {
+      hasher.isSuitable(passwordInfo) must beTrue
```
alexmojaki: Similarly include a test using both `hash` and `isSuitable`

silhouette-password-bcrypt/src/test/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasherSpec.scala
```diff
@@ -17,56 +17,82 @@ package com.mohiva.play.silhouette.password
...
-      val hasher2 = new BCryptPasswordHasher(2)
+    "return Some(true) if the stored log rounds are not equal the hasher log rounds" in new Context {
+      override val hasher = new BCryptPasswordHasher(5)
+
+      hasher.isDeprecated(passwordInfo) must beSome(true)
+    }
 
-      hasher1 == hasher2 must beFalse
+    "return Some(false) if the stored log rounds are equal the hasher log rounds" in new Context {
+      hasher.isDeprecated(passwordInfo) must beSome(false)
```
alexmojaki: Two tests using both `hash` and `isDeprecated`

silhouette/app/com/mohiva/play/silhouette/impl/providers/CredentialsProvider.scala
```diff
@@ -66,16 +64,16 @@ class CredentialsProvider @Inject() (
...
-              authInfoRepository.update(loginInfo, passwordHasher.hash(credentials.password)).map(_ => loginInfo)
+            if (passwordHasherRegistry.isDeprecated(hasher) || hasher.isDeprecated(authInfo).contains(true)) {
+              authInfoRepository.update(loginInfo, passwordHasherRegistry.current.hash(credentials.password)).map(_ => loginInfo)
             } else {
               Future.successful(loginInfo)
             }
           case Some(hasher) => throw new InvalidPasswordException(InvalidPassword.format(id))
           case None => throw new ConfigurationException(UnsupportedHasher.format(
-            id, authInfo.hasher, passwordHasherList.map(_.id).mkString(", ")
+            id, authInfo.hasher, passwordHasherRegistry.all.map(_.id).mkString(", ")
```
alexmojaki: I see you had to make the exact same changes twice. Keep in mind my suggestion to refactor the greater bulk of these two classes into one.

silhouette/test/com/mohiva/play/silhouette/impl/providers/BasicAuthProviderSpec.scala
```diff
@@ -135,15 +150,29 @@ class BasicAuthProviderSpec extends PlaySpecification with Mockito {
     lazy val fooHasher = {
       val h = mock[PasswordHasher]
       h.id returns "foo"
+      h.isSuitable(any) answers { p =>
```
alexmojaki: Does the mocking library make this necessary? The default implementation doesn't get used? It really doesn't know what the input type to `isSuitable` is? If you actually have to redefine it can I suggest `p.asInstanceOf[PasswordInfo].hasher == "foo"`?

silhouette/test/com/mohiva/play/silhouette/impl/providers/BasicAuthProviderSpec.scala
```diff
@@ -135,15 +150,29 @@ class BasicAuthProviderSpec extends PlaySpecification with Mockito {
...
     }
 
     /**
-     * An optional password hasher.
+     * A deprecated password hasher.
      */
     lazy val barHasher = {
       val h = mock[PasswordHasher]
       h.id returns "bar"
+      h.isSuitable(any) answers { p =>
```
alexmojaki: The only difference between the two hashers is the `id`, so can't you factor out a function?

silhouette/test/com/mohiva/play/silhouette/impl/providers/CredentialsProviderSpec.scala
```diff
@@ -107,15 +121,29 @@ class CredentialsProviderSpec extends PlaySpecification with Mockito {
     lazy val fooHasher = {
       val h = mock[PasswordHasher]
       h.id returns "foo"
+      h.isSuitable(any) answers { p =>
```
alexmojaki: These two hashers and the registry are exactly the same as in the `BasicAuthProviderSpec`, can't they be reused from a common location?

alexmojaki:
Hooray, no more metadata! Thanks for listening 😄

silhouette/test/com/mohiva/play/silhouette/impl/providers/BasicAuthProviderSpec.scala
akkie: I think this is a good idea.

silhouette-password-bcrypt/src/test/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasherSpec.scala
akkie: I think the best is to remove the line:
```
passwordInfo must beAnInstanceOf[PasswordInfo]
```

silhouette-password-bcrypt/src/test/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasherSpec.scala
akkie: For all the tests you mentioned, I can execute the `hash` method in the `Context` instead of using the hardcoded password info. What do you think?

silhouette/app/com/mohiva/play/silhouette/impl/providers/CredentialsProvider.scala
akkie: OK, I see how I can refactor this into a trait used by both implementations.

silhouette/test/com/mohiva/play/silhouette/impl/providers/BasicAuthProviderSpec.scala
akkie: Yes, the default implementation doesn't get used.

```
p.asInstanceOf[PasswordInfo].hasher == "foo"
```

Is much better. I'll change it.

silhouette/test/com/mohiva/play/silhouette/impl/providers/CredentialsProviderSpec.scala
akkie: OK, I'll see how I can refactor this out.

silhouette-password-bcrypt/src/test/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasherSpec.scala
alexmojaki: It's possible for both the `hash` and the `matches` method to change in such a way that they are still compatible, but `matches` is no longer compatible with previous hashes. Usually this kind of problem would be solved by comparing the output of `hash` with a hardcoded value, but since it's random we have to resort to double checking.

I suggest `val passwordInfos = Seq(hardcodedPasswordInfo, hasher.hash(password))` in the context and `for (passwordInfo <- passwordInfos)` in the tests.

Taking that further, it's also possible for the tests to pass by luck sometimes even if the methods are broken, and we can reduce that possibility by letting `passwordInfos` contain 5 hardcoded and 5 fresh values.

akkie: Iterating over a list in a test makes the test result incomprehensible in case of an error, because you do not see which value in the list is wrong. It shows only that a value is wrong, but not which value exactly. It does also not show that multiple values are wrong. This makes the code hard to debug. Instead we should write more test cases which cover all the possibilities.

alexmojaki: Surely it's easy to just add a `println` to see the troublesome value?

akkie: If the test runs in the CI system it's not. I think the test result should so meaningful that no further debugging, which values has failed, is necessary.

Or do you mean, add the `println` to the test in general?

```scala
"test" in new Scope {
  val list = Seq(true, true, false, false)

  for (v <- list) {
    println(v)
    v must beTrue
  }

}
```

Prints:
```
true
true
false
```

And not:
```
true
true
false
false
```

Because the test stops after the erroneous value.

akkie:
I've updated the pull request again with your suggestions.

silhouette-password-bcrypt/src/test/scala/com/mohiva/play/silhouette/password/BCryptPasswordHasherSpec.scala
```diff
@@ -17,56 +17,91 @@ package com.mohiva.play.silhouette.password
...
-      val hasher1 = new BCryptPasswordHasher(1)
-      val hasher2 = new BCryptPasswordHasher(1)
+  "The `isDeprecated` method" should {
+    "return None if the hasher isn't suitable for the given password info" in new Context {
+      val passwordInfo = PasswordInfo("scrypt", "")
+
+      hasher.isDeprecated(passwordInfo) must beNone
+    }
+
+    "return Some(true) if the stored log rounds are not equal the hasher log rounds" in new Context {
```
alexmojaki: Add a similar test for `isSuitable`: "return true when given a password info from a bcrypt hasher with different log rounds"

silhouette/app/com/mohiva/play/silhouette/api/util/PasswordHasher.scala
```diff
@@ -65,4 +62,71 @@ trait PasswordHasher {
...
+ *
+ * The password hasher registry defines the current password hasher which is able to hash all new passwords
+ * and also match the passwords store in the backing store for this algorithm. And also a list of deprecated
+ * hashers, which should match passwords that are stored in the baking store but which are different to the
+ * current hasher.
+ *
+ * @param current The current password hasher used by the application.
+ * @param deprecated The deprecated list of password hashers.
+ */
+case class PasswordHasherRegistry(current: PasswordHasher, deprecated: Seq[PasswordHasher] = Seq()) {
```
alexmojaki: Now that you have `PasswordProvider`, do you still need this class?

akkie: Yes, it makes sense, because it defines the usage of the password hashers application wide. You must define it only once and all password based providers get it injected. If we remove this class, then every password based provider must define the list of hashers again. We have one table that stores the password info and we have one definition of the hashers stored in this table.

akkie:
I've added the test. I hope now it's ready to merge :wink:

alexmojaki:
OK 👍

akkie:
Great :+1:

PR #466, opened 2016-04-12
AddingCasAuthInfoDAO

SBSMMO:
Required for seed project.

Also fixed a bug with correct username retrieval.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -35,18 +35,18 @@ object Build extends Build {
     base = file("silhouette-persistence"),
     dependencies = Seq(silhouette)
   )
-
-  val silhouettePersistenceMemory = Project(
-    id = "play-silhouette-persistence-memory",
-    base = file("silhouette-persistence-memory"),
-    dependencies = Seq(silhouettePersistence)
-  )
-
+  
   val silhouetteCAS = Project(
     id = "play-silhouette-cas",
     base = file("silhouette-cas"),
     dependencies = Seq(silhouette % "compile->compile;test->test")
   )
+
+  val silhouettePersistenceMemory = Project(
+    id = "play-silhouette-persistence-memory",
+    base = file("silhouette-persistence-memory"),
+    dependencies = Seq(silhouettePersistence, silhouetteCAS)
+  )  
 
   val silhouetteTestkit = Project(
     id = "play-silhouette-testkit",
--- a/project/BuildSettings.scala
+++ b/project/BuildSettings.scala
@@ -32,7 +32,7 @@ object BasicSettings extends AutoPlugin {
 
   override def projectSettings = Seq(
     organization := "com.mohiva",
-    version := "4.0.0-BETA4",
+    version := "4.0.0-BETA4-cas",
     resolvers ++= Dependencies.resolvers,
     scalaVersion := Dependencies.Versions.scalaVersion,
     crossScalaVersions := Dependencies.Versions.crossScala,
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -15,6 +15,7 @@
  */
 import sbt._
 
+
 object Dependencies {
 
   object Versions {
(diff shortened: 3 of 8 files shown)


akkie:
This would pull the CAS module if a user selects the persistence module, even if CAS isn't needed. I'm not sure what the best approach would be. Please give me some days.

But please could you submit a pull request, which fixes only the bug with username retrieval you have encountered.

arjunsol:
Perhaps if the persistence guice module could be overriden, or perhaps made private and explicitly included?

I could then build my own DelegableAuthInfoRepository in the seed project.

akkie:
I think we should remove the automatic binding for the persistence memory module. Then everyone can bind the DAO's as one likes. I would provide a default Guice module in every SBT module which serves as an example or which can be used standalone. What do you think?

arjunsol:
Is that to say that each Guice module would become private to each project, it would then be used thus:

install(new PersistenceModule())

?

akkie:
No, it's not private. I'm not sure if private makes sense, because then nobody can use it.

akkie:
I've removed the default binding #469

Now this pull request should be divided into two parts. The first part should fix the bug with the username retrieval you have encountered. The second should create a new module called `silhouette-cas-persistence-memory` which contains the persistence stuff. What do you think?

arjunsol:
Apologies I wasn't clear I mean a PrivateModule:

https://google.github.io/guice/api-docs/latest/javadoc/index.html?com/google/inject/PrivateModule.html

Unless I've misunderstood how this works.

akkie:
I think this is to confusing for the most of the users.

akkie:
@arjunsol Any news here? I plan to release an RC1 at the beginning of the next week, and it would be great if we can get this merged.

arjunsol:
I'll add the bug fix in a separate pull request today.

I have my doubts on creating a module to contain a few classes, for one it's not really scalable should other providers need to be added. It feels like it's a case for MultiBinder.

The memory module adds all default DAO's to the binder.
DelegableAuthInfoRepository then has Set[InMemoryAuthInfoDAO] injected in its constructor.

Any project using the persistence projects can add to the multibinder as needed.
Alternative the cas project could supply the DAO in the same fashion.

Thoughts?

akkie:
@arjunsol The MultiBinder sounds great!

arjunsol:
Yes I thought so, ok, I'll submit two pulls. One bug fix and one for swapping to MultiBinder. Haven't done a PoC yet but it looks like it would work.

akkie:
@arjunsol Any news? I'm ready to release the RC1!

arjunsol:
I've been looking at this today, having trouble getting the MultiBinder to work with the generics in DelegableAuthInfoDAO

I can get it to compile using:

`ScalaMultibinder.newSetBinder[DelegableAuthInfoDAO[_]](binder)`

However Guice then fails to inject, even if I use @Named

akkie:
Please let us remove the `provideAuthInfoRepository` method form the [PersistenceModule](https://github.com/mohiva/play-silhouette/blob/master/silhouette-persistence-memory/src/main/scala/com/mohiva/play/silhouette/persistence/memory/modules/PersistenceModule.scala#L56) and let us provide only the bindings for the DAOs. So the user can decide which DAOs he need and provide a binding for the `AuthInfoRepository` in the main module of the application.

The memory modules are anyway placeholder modules which gets replaced in real applications.

arjunsol:
OK, that makes sense.

akkie:
If you cannot find the time I could create the pull request, if this is OK for you! What do you think?

akkie:
See #477

arjunsol:
Apologies I have been away for the past week. I see you made the change, looks good. I'll keep an eye out for RC1 as I see the seed is using it. Then I'll add CAS to the seed project.

akkie:
I've removed all the in memory based modules. Please see #479 for a detailed explanation.

I've already released RC1 and I've also adapted the seed templates! An announcement comes in the next few days.

arjunsol:
Right you are, I'll add the CAS provider to the seed.

PR #470, opened 2016-05-03
[WIP] Prototyped refresh for OAuth2 Bearers

reactormonk:
# Pull Request Checklist
- [x] Have you read [How to write the perfect pull request](https://github.com/blog/1943-how-to-write-the-perfect-pull-request)?
- [x] Have you read through the [contributor guidelines](https://github.com/mohiva/play-silhouette/blob/master/CONTRIBUTING.md)?
- [ ] Have you [squashed your commits](https://docs.openshift.org/origin-m4/oo_contributors_guide.html#submitting-code)?
- [x] Have you added copyright headers to new files?
- [x] Have you suggest documentation edits?
- [ ] Have you added tests for any changed functionality?

`sbt` can't find `scripts/reformat`.
## Purpose

This PR adds refresh_token functionality.
## Background

I've tried to add some type safety via `OAuth2Bearer`, It might be a good idea to subclass `OAuth2Provider` in a similar manner, maybe adding the `OAuth2Info` via generic.
## References

https://tools.ietf.org/html/rfc6749#section-6
Supposedly you can link gitter chats, but I can't get the anchor.

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala
@@ -51,7 +51,16 @@ case class OAuth2Info(
   tokenType: Option[String] = None,
   expiresIn: Option[Int] = None,
   refreshToken: Option[String] = None,
-  params: Option[Map[String, String]] = None) extends AuthInfo
+  params: Option[Map[String, String]] = None,
+  generatedAt: Instant = new Instant()
+) extends AuthInfo {
+  def expired: Boolean = {
+    expiresIn match {
+      case Some(in) => generatedAt.plus((in * 1000).toLong).isAfter(new Instant())
+      case None     => false
+    }
+  }
+}
 
 /**
  * The Oauth2 info companion object.
--- a/silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala
@@ -184,28 +184,28 @@ trait OAuth2Provider extends SocialProvider with OAuth2Constants with Logger {
   }
 
   /**
-    * Refreshes the OAuth2Info token at the refreshURL. Example code:
-    *
-    * def bearer(token: String) = req.withHeaders("Authorization" -> s"Bearer ${token}")
-    *
-    * def executeOAuth(info: OAuth2Info, provider: OAuth2Provider) = {
-    *   val again = provider.refresh(info.refreshToken).flatMap(token =>
-    *     bearer(token, request).execute().map(_ -> token)
-    *   )
-    *   val response = if (info.expired) {
-    *     again
-    *   } else {
-    *     bearer(info, request).execute().flatMap({ resp =>
-    *       resp.status match {
-    *         case 401 => again
-    *         case _   => Future.successful((resp, info))
-    *       }
-    *     })
-    *   }
-    * }
-    *
-    * @param refreshToken The refresh token, as on OAuth2Info
-    */
+   * Refreshes the OAuth2Info token at the refreshURL. Example code:
+   *
+   * def bearer(token: String) = req.withHeaders("Authorization" -> s"Bearer ${token}")
+   *
+   * def executeOAuth(info: OAuth2Info, provider: OAuth2Provider) = {
+   *   val again = provider.refresh(info.refreshToken).flatMap(token =>
+   *     bearer(token, request).execute().map(_ -> token)
+   *   )
+   *   val response = if (info.expired) {
+   *     again
+   *   } else {
+   *     bearer(info, request).execute().flatMap({ resp =>
+   *       resp.status match {
+   *         case 401 => again
+   *         case _   => Future.successful((resp, info))
+   *       }
+   *     })
+   *   }
+   * }
+   *
+   * @param refreshToken The refresh token, as on OAuth2Info
+   */
 
   def refresh(refreshToken: String): Option[Future[OAuth2Info]] = {
     settings.refreshURL.map({ url =>
--- a/silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala
@@ -19,6 +19,9 @@
  */
 package com.mohiva.play.silhouette.impl.providers
 
+import javax.inject.Inject
+
+import com.mohiva.play.silhouette.api.util.Clock
 import java.net.URLEncoder._
 
 import com.mohiva.play.silhouette.api._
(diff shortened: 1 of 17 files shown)


silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala
```diff
@@ -165,6 +216,34 @@ trait OAuth2Provider extends SocialProvider with OAuth2Constants with Logger {
       info => Success(info)
     )
   }
+
+  def maybeRefresh(request: WSRequest, bearer: OAuth2Bearer): Future[(WSResponse, OAuth2Bearer)] = {
```
akkie: Please could you explain how this should be called from application code?

akkie:
The introduction of the `OAuth2Bearer` brings a lot of inconsistency to the library. Please be aware that the library doesn't handle OAuth2 only. The library is designed in a way that multiple protocols can be handled over a simple and well shaped interface. Now the `OAuth2Bearer` brings a lot of trouble. I think you stumbled over the issue with the DAOs.

The library is currently designed in the way that every provider can handle a single `AuthInfo` implementation. That cannot be easily changed and I see no reason why we should change that, because the refresh can be handled with `OAuth2Info` because it contains all needed values.

I'm also not satisfied with the introduction to the bearer term itself. I've described the reasons in a [mailing list post](https://groups.google.com/forum/#!searchin/play-silhouette/bearer/play-silhouette/9gYkHUybzuo/IsrsP1BiIAAJ) a while ago. Silhouette should be open to every transport method. Currently this is implemented with the `RequestExtractor`. The only missing part is the extraction from the the `Bearer` token scheme.

As noted in the chat, the provider should provide two additional methods:
```scala
def refresh[B]()(implicit request: ExtractableRequest[B]): Future[Either[Result, OAuth2Info]]
def refresh(refreshToken: String): Future[Either[Int, OAuth2Info]]
```

The first method can extract the token from different request parts. It then returns either the OAuth2Info or the status code on error. The second method can be used to retrieve the token without a request. This would follow the same semantic as currently implemented for the access_token retrieval.

Currently your implementation looks to me as it solves only your use-case. But please try to design the changes so that all possible use-cases can be handled.

reactormonk:
I can't implement something that depends on anything related to `Request` or `Result`, because these describe incoming HTTP requests. This PR would handle outgoing requests.

The bearer isn't a provider, it's a way to typecheck for `token_type` `"Bearer"`, that's why I switched over to ADT. Maybe I'd better go with a different solution that wouldn't involve ADTs.

silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala
reactormonk: val request = httpLayer.url("some/url/on/the/api").withMethod("post")
    val bearer = <from somewhere>
    val provider = <from somewhere>
    provider.maybeRefresh(request, bearer).map({ case (response, newBearer) =>
          dao.update(newBearer)
    })

akkie:
```scala
val request = httpLayer.url("some/url/on/the/api").withMethod("post")
val bearer = <from somewhere>
val provider = <from somewhere>
provider.maybeRefresh(request, bearer).map({ case (response, newBearer) =>
      dao.update(newBearer)
})
```

In this case you misuses the library as client for your own endpoint. But the provider should only deal with requests to the authentication endpoints. The name says it: The `OAuth2Provider` deals with provider(Facebook, Google, ...) related requests and describes the flow handled by the provider. Sorry, but this is something I do not integrate into the library. Such functionality could be implemented into a separate client component. But even this client component shouldn't be part of the core library. Normally you use a API specification like Swagger to describe your endpoints and generate a client for it.

The flow should be as follow. You describe a `refresh` endpoint in your application. In the action which handles this endpoint you call the `OAuth2Provider.refresh` method and pass the current request. The provider extracts the refresh token from request, then it handles the refresh flow against the authentication provider and gives you a `OAuth2Info` back. Now you can store the info or give it back to your client.

silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala
akkie: Please see my notes below.

reactormonk:
Ok, I'll reduce to `refresh` in that case.

    def refresh(refreshToken: String): Future[Either[Int, OAuth2Info]]
    
Why the Int? `Either[WSResponse, OAuth2Info]` sounds more useful to me.

    def refresh[B]()(implicit request: ExtractableRequest[B]): Future[Either[Result, OAuth2Info]]
    
I don't see why I would need this method, refreshing is active, it never happens on incoming requests according to the spec IIRC.

akkie:
> Why the Int? Either[WSResponse, OAuth2Info] sounds more useful to me.

The question is: Is it useful to return the status code from the authentication provider? I think it's not necessary to return the code because the logic regarding the return codes is handled inside the refresh method. What do you think?

Anyway, I do not like to declare the `WSResponse` to the interface because this type is part of the WS Play library which gets replaced in the near future.

> I don't see why I would need this method, refreshing is active, it never happens on incoming requests according to the spec IIRC.

It's not really needed, but it thought it would reduce boilerplate code for the flow I've described in my last post.

reactormonk:
> Anyway, I do not like to declare the WSResponse to the interface because this type is part of the WS Play library which gets replaced in the near future.
How about a `case class Response(status: Int, body: String)`? I think it would be helpful to have the response from the API.

> It's not really needed, but it thought it would reduce boilerplate code for the flow I've described in my last post.
That flow will never happen, refreshing tokens is something active, not via incoming request. Or are you talking about implementing refresh capabilities when your play app is an OAuth provider?

akkie:
> How about a case class Response(status: Int, body: String)? I think it would be helpful to have the response from the API.

What is body?

> That flow will never happen, refreshing tokens is something active, not via incoming request. Or are you talking about implementing refresh capabilities when your play app is an OAuth provider?

The app can be an API for mobile applications as example.

reactormonk:
> What is body?

The response from the server in case the token refresh fails.

> The app can be an API for mobile applications as example.

I can't see how that would work out, I've never done mobile applications.

akkie:
What about:
```scala
case class TokenRefreshException(msg: String, code: Int, cause: Throwable = null)
```

reactormonk:
Actually, I'm using `buildInfo` here, so I inherit the error handling from that method.

reactormonk:
How can I tell the tests not to look at `generatedAt`?

akkie:
> How can I tell the tests not to look at generatedAt?

The `generatedAt` property is not the problem because you can set it in test. The problem is your `expired` method because it instantiates the `Instant` class internally. You should pass a `Clock` instance to your `expired` method.

```scala
def expired(clock: Clock): Boolean
```

reactormonk:
`Clock`? http://www.joda.org/joda-time/apidocs/index.html You mean `Instant`?

akkie:
I meant https://github.com/mohiva/play-silhouette/blob/030608725c3f5f4abdf5dc171020a93c37a2a89b/silhouette/app/com/mohiva/play/silhouette/api/util/Clock.scala

reactormonk:
There we go.

reactormonk:
Could you help me a bit with the mocking? I've never used mockito before.

reactormonk:
Apparently `abstract class OAuth2Provider @Inject() (clock: Clock)` isn't the right way to go inject `Clock` - where should I put it in?

akkie:
Sorry, I'm away until Monday evening.

reactormonk:
Might consider https://docs.oracle.com/javase/8/docs/api/java/time/Clock.html too.

akkie:
Why do you have closed the pull request?

reactormonk:
Because I switched over to http4s, and I have no more interest in maintaining the PR.

reactormonk:
I'm not sure where to `@Inject()` `Clock`. I'd guess `infoReads` needs it, or make that via a separate case class, so `Clock` is only in `OAuth2Provider`?

akkie:
I think you do not need the `Clock` instance here because you can pass an `Instant` to the `expired` method. This is enough.

akkie:
Implemented with https://github.com/minutemen/silhouette/pull/22

PR #473, opened 2016-05-17
Fixing issue with CAS username

arjunsol:
## Purpose

Fixes an issue in the CasProvider where an incorrect LoginInfo object was being constructed.

## Background Context

The unit tests pass as they are contrived. However the previous code falls over at runtime.


akkie:
:+1:

PR #493, opened 2016-09-29
JWT Authenticator - accept different signing algorithms for JWT

yaronyam:
Refactored _JWTAuthenticator_ to accept different signing/verifying algorithms.
Added an optional implementation of ECDSA algorithm - _ECDSAAlgorithmFactory_

*Will add UT's for _ECDSAAlgorithmFactory_ is there's an interest

### DIFF
--- a/silhouette-testkit/app/com/mohiva/play/silhouette/test/Fakes.scala
+++ b/silhouette-testkit/app/com/mohiva/play/silhouette/test/Fakes.scala
@@ -23,6 +23,7 @@ import com.mohiva.play.silhouette.api.repositories.AuthenticatorRepository
 import com.mohiva.play.silhouette.api.services.{ AuthenticatorService, IdentityService }
 import com.mohiva.play.silhouette.api.util.Clock
 import com.mohiva.play.silhouette.impl.authenticators._
+import com.mohiva.play.silhouette.impl.authenticators.jwt.{JWTAuthenticator, JWTAuthenticatorService, JWTAuthenticatorSettings}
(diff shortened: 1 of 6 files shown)


akkie:
I currently implement a new JWT abstraction into the [framework agnostic version](https://github.com/mohiva/silhouette) of Silhouette. With this abstraction it's possible to use different JWT libraries. It's planned to use `atlassian-jwt` for backward compatibility and also [jwt-scala](https://github.com/pauldijou/jwt-scala) which supports all kind of signing algorithms. I'll ping you if I've commited the code. So then we can consider if we port this abstraction into this project and maybe add more signing algorithms to the `silhouette-jwt-atlassian` package. What do you think?

yaronyam:
I would remove dependency on Atlassian awt either way. The fewer dependencies the better. Especially when it itroduces transitive dependencies and a specialized resolver. It doesnt provide much functionality and its limiter to HMAC algorithm. Its only used in 1 method in this project and its interaces are not exposed externally, so I wouldn't worry too much about backward compatebility. I would use Nimbus awt soley and upgrade its version to 4.x which is very developer friendly - http://connect2id.com/products/nimbus-jose-jwt/examples

akkie:
@yaronyam I've finalized the [JWT implementation](https://github.com/minutemen/silhouette/pull/12) for the [framework agnostic version](https://github.com/minutemen/silhouette) of Silhouette. This implementation is based on jose4j instead on nimbus. It would be cool if you could find the time to review it :wink:

yaronyam:
looks good @akkie  :thumbsup:

riomus:
Any info on this?

akkie:
The framework agnostic version is still in development. Any contribution is highly appreciated.

PR #502, opened 2016-12-07
WIP - Basic concept check for user state param to and fro

### LINKED ISSUE #391: Add support for per-authentication-flow state to SocialProviders (feature request/design proposal)
The main application for this is redirection to the same page that a user originally requested but was not authenticated for; in OAuth2 there is a separate redirect_url (aka callbackURL) param, but most providers restrict that to a preset whitelist of URLs in order to combat CSRF attacks; the best practice recommended by e.g. [Google](https://developers.google.com/identity/protocols/OpenIDConnect#sendauthrequest) is to have `state` be a structured field that includes both the CSRF token _and_ "any other information needed to recover the context when the user returns to your application, e.g. the starting URL." Sessions and flash scope don't quite work, because they break in the case where a user has more than one tab/window open.

My proposal is to make two changes to the SocialProvider trait:
1. Add a `userState: Map[String, String]` param to `SocialProvider.authenticate[B]()`, which will be inserted into the auth redirect on the initial request (i.e. when `authenticate()` returns a `Left[Result]`) and be re-inserted into the authInfo in a provider-specific way on auth completion.
2. Add a `retrieveUserState(authInfo: A): Map[String, String]` which will allow a caller to retrieve this per-flow state on a successful authentication.

I have a fairly good grasp on how to implement this for OAuth2Provider - change `authenticate()` to combine both `serializedState` (which I'd probably rename to `csrfState` for clarity) and the new `userState` param to get the `state` auth URL parameter, and change the `StateProvider.validate[B]` method to take an explicit `providerState` parameter instead of reading out the whole `state` callback URL param.

I would appreciate both comments on this design as it applies to OAuth2 and tips for how to implement this for other providers; once I've gotten some design feedback I'll start putting a PR together.

Saheb:
# Pull Request Checklist

* [ ] Have you [squashed your commits](https://www.playframework.com/documentation/2.5.x/WorkingWithGit#Squashing-commits)?
* [ ] Have you added copyright headers to new files?
* [ ] Have you suggest documentation edits?

## Fixes

Fixes #391

## Purpose

Makes authenticate call with user params and returns them after the call is successful, helping application developers to redirect to the page the user came from. Still in a conceptual stage as I am not sending it to the auth provider like Github or facebook, just returning it back after the authentication passes through. We may want to send some way to send it to External Auth Providers if they have some way to pass it back to the callback url

## Background Context

Was simple and discussed on the issue thread

## References

Are there any relevant issues / PRs / mailing lists discussions?


Saheb:
@akkie - Just got the first version up with the basic flow, check it out and then we can improvise over it. 

Side note - I know there is code duplication, we can replace the old authenticate call with None as default value, but the clients who are using will have to change the way they process the result as the return type will change from Result to StatefulResult. Hence haven't removed the earlier method yet!

silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala
```diff
@@ -137,6 +137,43 @@ trait OAuth2Provider extends SocialProvider with OAuth2Constants with Logger {
     }
   }
 
+  override def authenticate[B](userStateParam: Option[Map[String, String]])(implicit request: ExtractableRequest[B]): Future[Either[StatefulResult, OAuth2Info]] = {
```
akkie: Why `Option`? If a user uses this method then he probably want's to send state. Also, the `Map` itself can be empty.

I think `Map[String, String]` is enough.

silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialStateProvider.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import com.mohiva.play.silhouette.api.util.ExtractableRequest
+import play.api.mvc.Result
+
+import scala.concurrent.Future
+
+/**
+ * Created by sahebmotiani on 29/11/2016.
+ */
+
```
akkie: This line can be removed.

silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialStateProvider.scala
```diff
@@ -0,0 +1,28 @@
+package com.mohiva.play.silhouette.impl.providers
```
akkie: The doc comment is missing.

silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialStateProvider.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import com.mohiva.play.silhouette.api.util.ExtractableRequest
+import play.api.mvc.Result
+
+import scala.concurrent.Future
+
+/**
+ * Created by sahebmotiani on 29/11/2016.
```
akkie: Please could you add a meaningful doc comment

silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialStateProvider.scala
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+case class StatefulResult(result: Result, userState: Option[Map[String, String]])
+
+trait SocialStateProvider extends SocialProvider {
```
akkie: The doc comment is missing.

silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialStateProvider.scala
```diff
@@ -0,0 +1,28 @@
...
+   * Authenticates the user and returns the auth information.
+   *
+   * Returns either a AuthInfo if all went OK or a Result that the controller sends
+   * to the browser (e.g.: in the case of OAuth where the user needs to be redirected to
+   * the service provider).
+   *
+   * @param request The request.
+   * @return Either a Result or the AuthInfo from the provider.
+   */
+  def authenticate[B](userStateParam: Option[Map[String, String]])(implicit request: ExtractableRequest[B]): Future[Either[StatefulResult, A]]
```
akkie: I would the variable name `userState` as like in the other signatures.

silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialStateProvider.scala
```diff
@@ -0,0 +1,28 @@
...
+
```
akkie: This line could be removed

akkie:
Some doc comment issues but overall it looks good. Yes the old authenticate method can call the new method with an empty map and it can extract the result from the `StatefulResult`. Now the `OAuth2StateProvider` must handle the user state.

silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala
Saheb: Agreed, my bad

silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialStateProvider.scala
Saheb: sure

silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialStateProvider.scala
Saheb: sure

silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialStateProvider.scala
Saheb: done

Saheb:
@akkie - Do we need to amend OAuthStateProvider to have userState, I am not clear as to why we need that?

akkie:
The user state must be sent to the provider somehow! How would you send it to the provider?

akkie:
@Saheb I'm not really sure if the current state provider is the right place to handle the user state, because it's meant for CSRF protection. I think we need a state provider, but in another form. The state provider should only sign and serialize/deserialize the state. The logic for the different state parts should be decoupled from this provider. So that we have a CSRFState handler and a UserState handler both with a unique interface. What do you think?

Saheb:
@akkie I was wondering what do we achieve by having a separate interface for userState? Just to make it more modular? And separation of responsibilities? Either way, in the end, we only have state param to send to the Social provider and that is going to come back. 

If we do plan to have it decoupled? How do we merge them back and send it as a value to state param?

akkie:
It was just an idea! The current state provider was only designed to handle the CSRF state and the interface uses methods directly created for this use case. So the idea was to decouple the different use-cases, so that it's possible to later handle different state parts if needed, without touching existing state handlers. Yes, the idea is it to make it more modular.

> Either way, in the end, we only have state param to send to the Social provider and that is going to come back.

In my opinion, the stats param is a composite of many different use-cases. Maybe a user do not need CSRF validation because it's handled by the client, but the user state is needed on the backend. This would not be possible with the current implementation.

> How do we merge them back and send it as a single value to state param?

This is the task of the state provider. It serializes the state from the different state handlers. Maybe a state handler returns a `Map[String, String]`. So the state providers combines those maps, serializes it and signs it.

Saheb:
Would placing the userState retriever in the SocialStateProvider be a good idea? How do we make the userState reach the backend, more importantly, how will the request handler on backend map the userState with the individual request (in case of multiple requests)? Am I making sense here?

akkie:
> Would placing the userState retriever in the SocialStateProvider be a good idea?

What do you understand by userState retriever? Do you mean the handler I mentioned in my last posts? If so, then I would place all the state related definitions into into the `com.mohiva.play.silhouette.impl.providers.oauth2.state` package. The `SocialStateProvider ` should be placed in the `SocialProvider.scala` file.

> How do we make the userState reach the backend, more importantly, how will the request handler on backend map the userState with the individual request (in case of multiple requests)?

Sorry, I don't really understand what you mean? Could you please explain this a bit more?

Saheb:
App will call authenticate on some provider. The provider will append state while hitting the auth url and then it will be redirected to redirect url. When that page will loaded/requested, it will append the state to it. Now how will the App retrieve userState from the state param?

akkie:
This is the task of the state provider. It gets the request passed, from which it can extract state from the state parameter. It checks the signature and if it's valid it deserializes the state value into a JSON object.

As example:
```json
{
  "userState": {
    "param1": "value1",
    "param2": "value2"
  },
  "csrfState": "some-id"
}
```

The different parts from the JSON will then be passed to the handlers, which do additional validations if needed. The state provider has methods which allow to check if a handler is available and to get a handler. So that the OAuth2 provider can get the extracted user state, from the user state handler, as example.

Saheb:
How about something like this?

```
trait OAuth2StateProvider{
...
def getState() : Option[State] = None
...
}
// The below ones in state package like the CookieState...
case class UserState(userState: Map[String, String])

class UserStateProvider extends OAuth2StateProvider 
```
And update methods of existing CookieStateProvider to use getUserState white creating state etc...

akkie:
@Saheb I would implement a more generic solution. With your proposal a user is not able to use an additional state handler with the `OAuth2StateProvider`, because it's not easily extendable.

Maybe something like this:
```scala
/**
 * Handles state for different purposes.
 */
trait StateHandler {

  /**
   * Validates the state.
   *
   * @param request The request instance to get additional data to validate against.
   * @param ec      The execution context to handle the asynchronous operations.
   * @tparam B The type of the request body.
   * @return True if the state is valid, false otherwise.
   */
  def validate[B](implicit request: ExtractableRequest[B], ec: ExecutionContext): Future[Boolean]

  /**
   * Gets the state params the handler can handle.
   *
   * @param ec The execution context to handle the asynchronous operations.
   * @return The state params the handler can handle.
   */
  def state(implicit ec: ExecutionContext): Future[Map[String, String]]
}

/**
 * A state handler which can publish its internal state to the client.
 *
 * Some state handlers, like the CSRF state handler, needs the ability to publish state to a cookie. So if you have
 * such a state handler, then mixin this trait, to publish the state to the client.
 */
trait ResultStateHandler {
  self: StateHandler =>

  /**
   * Publishes the state to the client.
   *
   * @param result  The result to send to the client.
   * @param request The current request.
   * @tparam B The type of the request body.
   */
  def publish[B](result: Result)(implicit request: ExtractableRequest[B]): Future[Result]
}

/**
 * Provides a way to handle different types of state for provides that allow a state param.
 *
 * Some authentication protocols defines a state param which can be used to transport some
 * state to an authentication provider. The authentication provider sends this state back
 * to the application, after the authentication to the provider was granted.
 *
 * The state parameter can be used for different things. Silhouette provides two state handlers
 * out of the box. One state handler can transport additional user state to the provider. This
 * could be an URL were the user should be redirected after authentication to the provider, or
 * any other per-authentication based state. An other important state handler protects the
 * application for CSRF attacks.
 */
trait StateProvider {

  /**
   * The concrete instance of the state provider.
   */
  type Self <: StateProvider

  /**
   * The states a provider can handle.
   */
  def states: Set[StateHandler]

  /**
   * Creates a copy of the state provider with a new handler added.
   *
   * There exists two types of state handlers. The first type are global state handlers which can be configured
   * by the user with the help of a configuration mechanism or through dependency injection. And there a local
   * state handlers which are provided by the application itself. This method exists to handle the last type of
   * state handlers, because it allows to extend the list of user defined state handlers from inside the library.
   *
   * @param handler The handler to add.
   * @return A new state provider with a new handler added.
   */
  def withHandler(handler: StateHandler): Self

  /**
   * Unserializes the state handlers from the state param.
   *
   * @param request The request to read the value of the state param from.
   * @param ec      The execution context to handle the asynchronous operations.
   * @tparam B The type of the request body.
   * @return The list of state handlers on success, an error on failure.
   */
  def unserialize[B](implicit request: ExtractableRequest[B], ec: ExecutionContext): Future[Set[StateHandler]]

  /**
   * Serializes the state handlers into a single state value which can be passed with the state param.
   *
   * @param ec The execution context to handle the asynchronous operations.
   * @return The serialized state as string.
   */
  def serialize(implicit ec: ExecutionContext): Future[String]
}

class CsrfStateHandler extends StateHandler with ResultStateHandler
class UserStateHandler extends StateHandler
```
Is it self-explaining? If not, please ask!

Saheb:
This is self-explanatory @akkie. I wonder why I couldn't come up with something more generic like this, let me work on it and get back with a solution.

Saheb:
@akkie - I have come up with something, it's still in progress, but if you can have a look and let me know if the direction is correct or not?
I am confused with the publishing trait and how is the internal state of a state handler important to the client? As we will be using the state of all handlers and generating the stateParam, shouldn't that be published to the client instead? Point being, at the StateProvider level and not at handler level?

akkie:
Hi @Saheb, this is not really I head in mind. The state provider does only handle the serialization/deserialization logic. 

The serialize method takes the internal list of state handlers(`Set[StateHandler]`), then it iterates over this list and calls the `state` method of the state handler. After that the serialize method should have a structure like:

```scala
Map[String, Map[String, String]](
  "UserState" -> Map(
    "url" -> "http://my.url.com"
  ),
  "CsrfState" -> Map(
    "token" -> "long.token"
  )
)
```

The key of the `Map` is the class name of the state handler. Now the serialize method takes this map and serializes it to a string. You can convert the Map to a JSON instance to serialize it. I think this is easier.

The unserialize method takes the string and deserializes it into the above map. Now it iterates over the state handlers as well and sets the state to every handler. Maybe we should add a `fromState` method to the `StateHandler` which instantiates a new state handler from the given state. On the new instance we now call the `validate` method of the state handler to validate the internal state.

The handler itself should not deal with serialization/deserialization logic, because this is only needed once at the provider level. The internal state of a handler is always represent with a `Map[String, String]`.

The publish method is used to store the ID, which is generated by the CSRF handler and transported in the state param, also in the cookie, so that after authentication on the social provider, the CSRF state handler can validate the ID decoded from the state param with the ID stored in the cookie. The user state handler does not need such logic. So only these handlers which need to store some additional data on the client can implement this method. The social provider can then get the list of state handlers from the state provider and call the publish method on every handler which has implemented the `PublishableStateHandler` trait.

I think the `StateHandler`, the `PublishableStateHandler` and the `StateProvider` should be prefixed with `Social` and be placed in the `SocialProvider` trait. The state handler implementations can be placed in the `com.mohiva.play.silhouette.impl.providers.state` package. Another possibility would it to place the stzate based traits also in the  `com.mohiva.play.silhouette.impl.providers.state` package.

I hope now it's a bit more clear :wink:

Saheb:
You caught it right, Serialization/Deserialization was happening multiple times, I was extracting it again at handler using the request, which isn't needed. Doing it once at StateProvider would solve that issue and then we can use the returned value. 

From the StateProviderImpl point of view, I am already doing serialization as your described. Not sure if you noticed that file. Fancy a better name for this impl? Suggestions?
https://github.com/mohiva/play-silhouette/pull/502/files#diff-d391c64051043f7fd1fd48d30cbf9304

```scala
override def serialize(implicit ec: ExecutionContext): Future[String] = {
    handlers.foldLeft(Future(Map.empty[String, Map[String, String]]): Future[Map[String, Map[String, String]]])((a, k) => {
      for {
        aParam <- a
        kParam <- k.state
      } yield (aParam + (k.toString -> kParam))
    }).map(state => Base64.encode(Json.toJson(state)))
  }
```
Base64 encoding is good enough for state param?

I will update the unserialize to take the Map instead of deserializing it again.

Thanks for the details explanation, things are clear now.

akkie:
:+1: I think `StateProviderImpl` is OK! Base64 is also OK!

Saheb:
@akkie - have come with something, as well as some tests to verify the flow. Also a clone of OAuth2ProviderUpated, which has the new authenticate implementation(There is a discussion point in that file(https://github.com/mohiva/play-silhouette/pull/502/commits/f7f068f9bdb7c00766c4bd17e4a828019c89551c#diff-bcce5d847036156700e4d684fd180476)). Let me know what you think and how we can improve it.

akkie:
@Saheb I had some ideas how we can improve the implementation with a more type safe approach. I'll create a base implementation so that we can discuss it.

akkie:
@Saheb I've pushed a base implementation of my ideas. The main parts are located in the [SocialStateProvider.scala](https://github.com/akkie/play-silhouette/blob/social-state-handler/silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialStateProvider.scala) file, in the [OAuth2Provider](https://github.com/akkie/play-silhouette/blob/social-state-handler/silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala#L76) class and in the [UserStateItemHandler.class](https://github.com/akkie/play-silhouette/blob/social-state-handler/silhouette/app/com/mohiva/play/silhouette/impl/providers/state/UserStateItemHandler.scala) file.

Maybe the naming of some components could be improved as also the error handling. The idea was that a Map[String, String] is not really type-safe. With this implementation every item state handler can use every data structure. The user state is also implemented to use every data structure. It must only be serialized to JSON with the help of the Play JSON format.

Any suggestions, what do you think?

Saheb:
@akkie Overall it looks good, but I'll need more time to clone it and think more about it.

It looks much cleaner and safer with a generic JsValue being passed around, instead of constraining state to Map. And also liked the way you merged AuthInfo and UserState together. 👍 

I am not clear with the aspect of how versions are going to be handled (generating UUIDs of some sort I believe). Have you thought it through already? Let me know

I will try to clone your version and see if I have any more comments or anomalies. Let me know how you want to move forward with further implementation and testing. Happy to help in any way or complete your version.

akkie:
@Saheb :+1:  It would be create if you could help to complete my version, so that I can concentrate on the further Silhouette agnostic development.

Versions should be simple numeric values. For the user state handler this isn't relevant, because it can already handle every data type. But for other providers this could be relevant if the underlying item type changes. We should also consider that the state isn't really long living. So I'm not sure it makes generally sense to implement versioning.

Let me know what you think!

Saheb:
@akkie Okay, I will take you version as base and work from there, will update the PR soon.

Will let you know if I find some implementation issues.

akkie:
@Saheb Why do you revert the "Change to new domains" pull request? Maybe we should start a new pull request with exactly the changes I've made as base. I think otherwise it would be confusing with all the comments and changed files.

Saheb:
Apologies for that confusion. That was a mistake, I am going to start a new
PR.

Saheb:
#503 is the new PR

PR #503, opened 2017-01-26
Social State Handlers

### LINKED ISSUE #391: Add support for per-authentication-flow state to SocialProviders (feature request/design proposal)

Saheb:
# Pull Request Checklist

* [ ] Have you [squashed your commits](https://www.playframework.com/documentation/2.5.x/WorkingWithGit#Squashing-commits)?
* [x] Have you added copyright headers to new files?
* [x] Have you suggest documentation edits?

## Fixes

Fixes #391 

## Purpose

What does this PR do?

## Background Context

Why did you take this approach?

## References

#502

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
@@ -1,3 +1,18 @@
+/**
+ * Copyright 2015 Mohiva Organisation (license at mohiva dot com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 package com.mohiva.play.silhouette.impl.providers.state
 
 import javax.inject.Inject
--- a/silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
@@ -35,7 +35,7 @@ class CsrfStateItemHandler @Inject() (
    */
   override def item(implicit ec: ExecutionContext): Future[Item] = {
     idGenerator.generate.map { id =>
-      CsrfState(cookieSigner.sign(id))
+      CsrfState(id)
     }
   }
 
@@ -121,7 +121,7 @@ class CsrfStateItemHandler @Inject() (
    */
   private def clientState(implicit request: RequestHeader): Try[JsValue] = {
     request.cookies.get(settings.cookieName) match {
-      case Some(cookie) => cookieSigner.extract(cookie.value).map(Json.toJson(_))
+      case Some(cookie) => Try(Json.toJson(cookie.value))
       case None         => Failure(new OAuth2StateException(ClientStateDoesNotExists.format(settings.cookieName)))
     }
   }
--- a/silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -17,18 +17,18 @@ package com.mohiva.play.silhouette.impl.providers.state
 
 import javax.inject.Inject
 
-import com.mohiva.play.silhouette.api.util.{ ExtractableRequest, IDGenerator }
-import com.mohiva.play.silhouette.impl.providers.SocialStateItem.ItemStructure
-import com.mohiva.play.silhouette.impl.providers.{ PublishableSocialStateItemHandler, SocialStateItem, SocialStateItemHandler }
-import play.api.libs.json.{ Format, Json }
-import CsrfStateItemHandler._
 import com.mohiva.play.silhouette.api.crypto.CookieSigner
+import com.mohiva.play.silhouette.api.util.{ExtractableRequest, IDGenerator}
 import com.mohiva.play.silhouette.impl.exceptions.OAuth2StateException
-import play.api.mvc.{ Cookie, RequestHeader, Result }
+import com.mohiva.play.silhouette.impl.providers.SocialStateItem.ItemStructure
+import com.mohiva.play.silhouette.impl.providers.state.CsrfStateItemHandler._
+import com.mohiva.play.silhouette.impl.providers.{PublishableSocialStateItemHandler, SocialStateItem, SocialStateItemHandler}
+import play.api.libs.json.{Format, Json}
+import play.api.mvc.{Cookie, RequestHeader, Result}
 
 import scala.concurrent.duration._
-import scala.concurrent.{ ExecutionContext, Future }
-import scala.util.{ Failure, Success, Try }
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success, Try}
 
 /**
  * Csrf State is a sub type of SocialStateItem
@@ -40,8 +40,8 @@ case class CsrfState(value: String) extends SocialStateItem
 /**
  * Handles csrf state.
  *
- * @param settings The state settings.
- * @param idGenerator The ID generator used to create the state value.
+ * @param settings     The state settings.
+ * @param idGenerator  The ID generator used to create the state value.
  * @param cookieSigner The cookie signer implementation.
  */
 class CsrfStateItemHandler @Inject() (
@@ -61,7 +61,9 @@ class CsrfStateItemHandler @Inject() (
    * @return The state params the handler can handle.
    */
   override def item(implicit ec: ExecutionContext): Future[Item] = {
-    idGenerator.generate.map { CsrfState(_) }
+    idGenerator.generate.map {
+      CsrfState(_)
+    }
   }
 
   /**
@@ -75,7 +77,7 @@ class CsrfStateItemHandler @Inject() (
    */
   override def canHandle(item: SocialStateItem): Option[Item] = item match {
     case i: Item => Some(i)
-    case _       => None
+    case _ => None
   }
 
   /**
@@ -84,15 +86,16 @@ class CsrfStateItemHandler @Inject() (
    * This method should check if the [[unserialize]] method of this handler can unserialize the given
    * serialized state item.
    *
-   * @param item The item to check for.
+   * @param item    The item to check for.
    * @param request The request instance to get additional data to validate against.
+   * @tparam B The type of the request body.
    * @return True if the handler can handle the given state item, false otherwise.
    */
   override def canHandle[B](item: ItemStructure)(implicit request: ExtractableRequest[B]): Boolean = {
     item.id == ID && {
       clientState match {
         case Success(token) => token == item.data.as[Item]
-        case Failure(_)     => false
+        case Failure(_) => false
       }
     }
   }
@@ -147,7 +150,7 @@ class CsrfStateItemHandler @Inject() (
   private def clientState(implicit request: RequestHeader): Try[Item] = {
     request.cookies.get(settings.cookieName) match {
       case Some(cookie) => cookieSigner.extract(cookie.value).map(token => CsrfState(token))
-      case None         => Failure(new OAuth2StateException(ClientStateDoesNotExists.format(settings.cookieName)))
+      case None => Failure(new OAuth2StateException(ClientStateDoesNotExists.format(settings.cookieName)))
     }
   }
 }
@@ -178,10 +181,10 @@ object CsrfStateItemHandler {
 /**
  * The settings for the Csrf State.
  *
- * @param cookieName The cookie name.
- * @param cookiePath The cookie path.
- * @param cookieDomain The cookie domain.
- * @param secureCookie Whether this cookie is secured, sent only for HTTPS requests.
+ * @param cookieName     The cookie name.
+ * @param cookiePath     The cookie path.
+ * @param cookieDomain   The cookie domain.
+ * @param secureCookie   Whether this cookie is secured, sent only for HTTPS requests.
  * @param httpOnlyCookie Whether this cookie is HTTP only, i.e. not accessible from client-side JavaScript code.
  * @param expirationTime State expiration. Defaults to 5 minutes which provides sufficient time to log in, but
  *                       not too much. This is a balance between convenience and security.
--- a/silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
@@ -1,0 +1,172 @@
+package com.mohiva.play.silhouette.impl.providers.state
+
+import javax.inject.Inject
+
+import com.mohiva.play.silhouette.api.util.{ ExtractableRequest, IDGenerator }
+import com.mohiva.play.silhouette.impl.providers.SocialStateItem.ItemStructure
+import com.mohiva.play.silhouette.impl.providers.{ PublishableSocialStateItemHandler, SocialStateItem, SocialStateItemHandler }
+import play.api.libs.json.{ Format, JsValue, Json }
+import CsrfStateItemHandler._
+import com.mohiva.play.silhouette.api.crypto.CookieSigner
+import com.mohiva.play.silhouette.impl.exceptions.OAuth2StateException
+import play.api.mvc.{ Cookie, RequestHeader, Result }
+
+import scala.concurrent.duration._
+import scala.concurrent.{ ExecutionContext, Future }
+import scala.reflect.ClassTag
+import scala.util.{ Failure, Success, Try }
+
+class CsrfStateItemHandler[S <: SocialStateItem] @Inject() (
+  settings: CsrfStateSettings,
+  idGenerator: IDGenerator,
+  cookieSigner: CookieSigner)(
+  implicit
+  format: Format[S],
+  classTag: ClassTag[S]) extends SocialStateItemHandler with PublishableSocialStateItemHandler {
+
+  /**
+   * The version of Csrf State it can handle
+   */
+  private val Version = java.util.UUID.randomUUID().toString
+
+  /**
+   * The item the handler can handle.
+   */
+  override type Item = S
+
+  /**
+   * Gets the state item the handler can handle.
+   *
+   * @param ec The execution context to handle the asynchronous operations.
+   * @return The state params the handler can handle.
+   */
+  override def item(implicit ec: ExecutionContext): Future[Item] = {
+    idGenerator.generate.map { id =>
+      Json.toJson(cookieSigner.sign(id)).as[Item]
+    }
+  }
+
+  /**
+   * Indicates if a handler can handle the given [[SocialStateItem]].
+   *
+   * This method should check if the [[serialize]] method of this handler can serialize the given
+   * unserialized state item.
+   *
+   * @param item The item to check for.
+   * @return [[Some]] casted state item if the handler can handle the given state item, [[None]] otherwise.
+   */
+  override def canHandle(item: SocialStateItem): Option[Item] = item match {
+    case i: Item => Some(i)
+    case _       => None
+  }
+
+  /**
+   * Indicates if a handler can handle the given unserialized state item.
+   *
+   * This method should check if the [[unserialize]] method of this handler can unserialize the given
+   * serialized state item.
+   *
+   * @param item The item to check for.
+   * @param request The request instance to get additional data to validate against.
+   * @return True if the handler can handle the given state item, false otherwise.
+   */
+  override def canHandle[B](item: ItemStructure)(implicit request: ExtractableRequest[B]): Boolean = {
+    item.id == ID && {
+      clientState match {
+        case Success(token) => token == item.data
+        case Failure(_)    => false
+      }
+    }
+  }
+
+  /**
+   * Returns a serialized value of the state item.
+   *
+   * @param item The state item to serialize.
+   * @return The serialized state item.
+   */
+  override def serialize(item: Item): ItemStructure = ItemStructure(ID, Version, Json.toJson(item))
+
+  /**
+   * Unserializes the state item.
+   *
+   * @param item    The state item to unserialize.
+   * @param request The request instance to get additional data to validate against.
+   * @param ec      The execution context to handle the asynchronous operations.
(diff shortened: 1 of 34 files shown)


akkie:
@Saheb Any progress?

Saheb:
Sorry, I have not been able to make time to complete this one, I think this week I will get some time. If that doesn't work out, you might wanna think of other options. Will keep you updated.

akkie:
@Saheb No problem, if you do not find the time to polish the PR, I can undertake the task.

Saheb:
Give me this week, then you can take over @akkie

akkie:
:+1:

Saheb:
Have added CsrfStateItemHandler and have modified some method signatures. I would like to have your comment on the same.

Not sure how we can use Versioning and having them makes sense. Can you throw some light there? 

Once the implementation looks clean and acceptable, then I would work on updating and adding more unit tests.

akkie:
@Saheb Looks good to me :+1: 

The version was thought as a version for different implementations. The current implementation is version 1. If we change the algorithm then we could increase the version to 2 but keep also the old implementation. So a serialized state versioned with version 1 can be handled as also serialized state versioned with version 2.

But as stated in the other thread. I'm not sure this makes sense, because the state is really short living. So I think we should omit the version from state.

Saheb:
@akkie - Have fixed the CsrfState and updated some signatures, just wanted your early feedback before writing more tests and good to be on the same page. 

Have added a basic test to get some clarity. Check out the way User State might be used by library users.

akkie:
@Saheb I'm away from home over the weekend. Will check your changes next week.

> Check out the way User State might be used by library users. i.e. They will have to define a type/case class which is a subtype of SocialStateItem and won't easily be able to pass Map directly. It's a fair tradeoff to keep the state generic I guess. Right?

We could provide a `MapStateItem` as example, which can transform a `Map[String, String]` into state.

akkie:
@Saheb Looks good to me!

One thing that I'm not really sure is the signing of the state. I think the complete state should be signed instead of only the CSRF state item.

Saheb:
I have made the change which signs the aggregated state instead of just CsrfState, but I am not sure about the Csrf Token generation best practices. The Csrf State which is published to the cookie is mandatory to be signed or just randomly generated id is good from security standard perspective?

akkie:
The state value in the cookie must also be signed. The state itself can be signed differently from the cookie. This is not a problem.

silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
```diff
@@ -35,7 +35,7 @@ class CsrfStateItemHandler @Inject() (
    */
   override def item(implicit ec: ExecutionContext): Future[Item] = {
     idGenerator.generate.map { id =>
-      CsrfState(id)
+      CsrfState(cookieSigner.sign(id))
```
akkie: This is not needed here, otherwise we sign the ID and then the complete state. The signing should be moved to the cookie creation,

silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
```diff
@@ -0,0 +1,169 @@
...
+   * @param item    The item to publish.
+   * @param result  The result to send to the client.
+   * @param request The current request.
+   * @tparam B The type of the request body.
+   * @return The result to send to the client.
+   */
+  override def publish[B](item: Item, result: Result)(implicit request: ExtractableRequest[B]): Result = {
+    result.withCookies(Cookie(
+      name = settings.cookieName,
+      value = item.toString,
```
akkie: Signing should occur here.

```
value = cookieSigner.sign(item.toString),
```

Saheb: Sure 👍

silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
Saheb: That makes, hence I was confused about signing it multiple times. Thanks for catching it.

Saheb:
@akkie - how does it look now?

silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
```diff
@@ -0,0 +1,167 @@
+package com.mohiva.play.silhouette.impl.providers.state
```
akkie: The file level doc block is missing

silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
```diff
@@ -0,0 +1,167 @@
...
+import CsrfStateItemHandler._
+import com.mohiva.play.silhouette.api.crypto.CookieSigner
+import com.mohiva.play.silhouette.impl.exceptions.OAuth2StateException
+import play.api.mvc.{ Cookie, RequestHeader, Result }
+
+import scala.concurrent.duration._
+import scala.concurrent.{ ExecutionContext, Future }
+import scala.util.{ Failure, Success, Try }
+
+case class CsrfState(value: String) extends SocialStateItem
```
akkie: The class level doc block is missing

silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
```diff
@@ -0,0 +1,167 @@
...
+
+class CsrfStateItemHandler @Inject() (
```
akkie: The class level doc block is missing

silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
```diff
@@ -0,0 +1,167 @@
...
+    }
+  }
+}
+
+/**
+ * The companion object.
+ */
+object CsrfStateItemHandler {
+  val ID = "csrf-state"
+  /**
```
akkie: There should be a newline before a doc comment

silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
```diff
@@ -0,0 +1,167 @@
...
+   * Json Format for the Csrf State
+   */
+  implicit val csrfFormat: Format[CsrfState] = Json.format[CsrfState]
+}
+
+/**
+ * The settings for the Csrf State.
+ *
+ * @param cookieName The cookie name.
```
akkie: We should be consistent with the alignment of the multi param description. For new code we use a one line alignment. Please see the above doc comments.

silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
```diff
@@ -0,0 +1,167 @@
...
+    case _       => None
+  }
+
+  /**
+   * Indicates if a handler can handle the given unserialized state item.
+   *
+   * This method should check if the [[unserialize]] method of this handler can unserialize the given
+   * serialized state item.
+   *
+   * @param item The item to check for.
```
akkie: We should be consistent with the alignment of the multi param description. For new code we use a one line alignment. Please see the below doc comments.

silhouette/app/com/mohiva/play/silhouette/impl/providers/state/UserStateItemHandler.scala
```diff
@@ -0,0 +1,112 @@
+/**
```
akkie: This file has also the alignment issues

silhouette/test/com/mohiva/play/silhouette/impl/providers/SocialStateProviderSpec.scala
```diff
@@ -0,0 +1,36 @@
+package com.mohiva.play.silhouette.impl.providers
```
akkie: The doc blocks are missing

silhouette/test/com/mohiva/play/silhouette/impl/providers/oauth2/GitHubProviderSpec.scala
```diff
@@ -165,6 +186,11 @@ class GitHubProviderSpec extends OAuth2ProviderSpec {
     override lazy val oAuthInfo = Helper.loadJson("providers/oauth2/github.access.token.json")
 
     /**
+     * The statefulAuthInfo
```
akkie: Such declarations must not really be documented, because no documentation will be created for test code. I nevertheless prefer the documentation, because I find it more readable, but this is only a personal preference and I can also live without the documentation. But then it should be consistent with the rest of the file. This means, if a documentation is available for a variable declaration, then it should be also available for the rest of the declarations.

So if you use the documentation, then this should be a real sentence.

```
The stateful auth info.
```

silhouette/test/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandlerSpec.scala
```diff
@@ -0,0 +1,118 @@
+package com.mohiva.play.silhouette.impl.providers.state
```
akkie: The file and class level doc comments are missing

silhouette/test/com/mohiva/play/silhouette/impl/providers/state/DefaultSocialStateHandlerSpec.scala
```diff
@@ -0,0 +1,111 @@
+package com.mohiva.play.silhouette.impl.providers.state
```
akkie: The file and class level doc comments are missing. It has also the same formatting issues as the above file.

silhouette/test/com/mohiva/play/silhouette/impl/providers/state/UserStateItemHandlerSpec.scala
```diff
@@ -0,0 +1,73 @@
+package com.mohiva.play.silhouette.impl.providers.state
```
akkie: The file and class level doc comments are missing. It has also the same formatting issues as the above file.

silhouette/test/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandlerSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+import scala.util.Success
+import scala.concurrent.duration._
+import play.api.libs.concurrent.Execution.Implicits._
+import play.api.mvc.Cookie
+
+import scala.concurrent.Future
+
+class CsrfStateItemHandlerSpec extends PlaySpecification with Mockito with JsonMatchers {
+
+  "item method of CsrfStateItemHandler" should {
```
akkie: We should be consistent with the rest of the tests.

We write:
```
"The `item` method" should {
```

The `CsrfStateItemHandler` is not needed here because the file test only one class and therefore it's redundant.

Tests for the same method should be grouped:
```
"The `canHandle` method" should {
  "should return `None` if it can't handle the given `SocialStateItem`" in new Context {
    ...
  }

  "should return true if it can handle the given `ItemStructure`" in new Context {
    ...
  }
}
```
We do not use a point at the end of the sentence.

silhouette/test/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandlerSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+    "unserialize ItemStructure to CsrfState" in new Context {
+      implicit val request = FakeRequest()
+      await(csrfStateHandler.unserialize(itemStructure)) must beAnInstanceOf[CsrfState]
+    }
+  }
+
+  trait Context extends Scope {
+
+    import CsrfStateItemHandler._
+    /**
```
akkie: There should be a newline before the doc comment.

silhouette/test/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandlerSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+     * The ID generator implementation.
+     */
+    lazy val idGenerator = mock[IDGenerator].smart
+    /**
```
akkie: There should be a newline before the doc comment.

silhouette/test/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandlerSpec.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  "unserialize method of CsrfStateItemHandler" should {
+    "unserialize ItemStructure to CsrfState" in new Context {
+      implicit val request = FakeRequest()
+      await(csrfStateHandler.unserialize(itemStructure)) must beAnInstanceOf[CsrfState]
+    }
+  }
+
+  trait Context extends Scope {
+
```
akkie: This newline can be replaced.

silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
Saheb: Can you give an example of this one? I am not clear with this one.

silhouette/test/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandlerSpec.scala
Saheb: Done

silhouette/test/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandlerSpec.scala
Saheb: Done

silhouette/test/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandlerSpec.scala
Saheb: Done

silhouette/test/com/mohiva/play/silhouette/impl/providers/oauth2/GitHubProviderSpec.scala
Saheb: Sure, done.

silhouette/test/com/mohiva/play/silhouette/impl/providers/state/UserStateItemHandlerSpec.scala
Saheb: Done

silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
Saheb: Done

silhouette/test/com/mohiva/play/silhouette/impl/providers/SocialStateProviderSpec.scala
Saheb: Done

silhouette/test/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandlerSpec.scala
Saheb: Done

silhouette/test/com/mohiva/play/silhouette/impl/providers/state/DefaultSocialStateHandlerSpec.scala
Saheb: Done

akkie:
Cool :+1: Looking forward for the last changes.

silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
Saheb: @akkie - Need your input here?

akkie: @Saheb Sorry, I've not seen your comment.

The param description should be aligned in one line rather than with only using one space after the param name and the description.

The correct way would be to search the longest param name. In this case it's request. Then use one space and write the description. Then align all other descriptions, so that they are aligned in one vertical line.
```
* @param item    The item to publish.
* @param result  The result to send to the client.
* @param request The current request.
```

Instead of:
```
* @param item The item to publish.
* @param result The result to send to the client.
* @param request The current request.
```

A description starts also with an uppercase letter and ends with a dot. It's a real sentence. It's the same as with normal doc comments, they must end with a dot.

Saheb: @akkie The same rule doesn't hold for type params? Inferring from the way you indented in SocialStateProvider.

```scala
  /**
   * Publishes the state to the client.
   *
   * @param item    The item to publish.
   * @param result  The result to send to the client.
   * @param request The current request.
   * @tparam B The type of the request body.
```
How about the order between type params and params? Which comes first?

```scala
  /**
   * Authenticates the user and returns the auth information and the user state.
   *
   * Returns either a [[StatefulAuthInfo]] if all went OK or a [[play.api.mvc.Result]] that the controller
   * sends to the browser (e.g.: in the case of OAuth where the user needs to be redirected to the service
   * provider).
   *
   * @tparam S       The type of the user state item.
   * @tparam B       The type of the request body.
   * @param format   The JSON format to the transform the user state into JSON.
   * @param request  The request.
   * @param classTag The class tag for the user state item.
   * @return Either a [[play.api.mvc.Result]] or the [[StatefulAuthInfo]] from the provider.
```

Saheb:
@akkie by last changes did you meant the formatting changes I am working on? Or would you like the test replicated for all OAuth2Providers, like the way I have modified GithubProviderSpec?

silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
akkie: "Align parameters descriptions" is the default setting in IntelliJ IDEA. There is no such settings for type params. Currently I do not align type params because mostly they have the same length(A, B, C) if I use multiple of them. And currently I have params first and then type params.

```scala
/**
   * Authenticates the user and returns the auth information and the user state.
   *
   * Returns either a [[StatefulAuthInfo]] if all went OK or a [[play.api.mvc.Result]] that the controller
   * sends to the browser (e.g.: in the case of OAuth where the user needs to be redirected to the service
   * provider).
   *
   * @param format   The JSON format to the transform the user state into JSON.
   * @param request  The request.
   * @param classTag The class tag for the user state item.
   * @tparam S The type of the user state item.
   * @tparam B The type of the request body.
   * @return Either a [[play.api.mvc.Result]] or the [[StatefulAuthInfo]] from the provider.
```

I think I should document this in the contribution guide.

akkie:
I meant only the formatting changes. But thanks that you say it, the tests should be replicated because we do this already for the other tests. I haven't noticed that before.

Saheb:
```bash
[info] Main Scala API documentation to /Users/sahebmotiani/github/play-silhouette/silhouette-testkit/target/scala-2.11/api...
model contains 14 documentable templates
model contains 12 documentable templates
model contains 19 documentable templates
[info] Main Scala API documentation successful.
[info] Main Scala API documentation successful.
[info] Main Scala API documentation successful.
model contains 19 documentable templates
[info] Main Scala API documentation successful.
[info] Main Scala API documentation successful.
model contains 322 documentable templates
[error] /Users/sahebmotiani/github/play-silhouette/silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialStateProvider.scala:38: Could not find any member to link for "Set".
[error] /**
[error] ^
[error] /Users/sahebmotiani/github/play-silhouette/silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialStateProvider.scala:28: Could not find any member to link for "AuthInfo".
[error] /**
[error] ^
[error] /Users/sahebmotiani/github/play-silhouette/silhouette/app/com/mohiva/play/silhouette/impl/providers/state/UserStateItemHandler.scala:55: Could not find any member to link for "Some".
[error]   /**
[error]   ^
[error] /Users/sahebmotiani/github/play-silhouette/silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala:69: Could not find any member to link for "Some".
[error]   /**
[error]   ^
[error] /Users/sahebmotiani/github/play-silhouette/silhouette/app/com/mohiva/play/silhouette/impl/providers/OAuth2Provider.scala:113: Could not find any member to link for "play.api.mvc.Result".
[error]   /**
[error]   ^
[error] /Users/sahebmotiani/github/play-silhouette/silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialStateProvider.scala:315: Could not find any member to link for "Some".
[error]   /**
[error]   ^
[error] /Users/sahebmotiani/github/play-silhouette/silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialStateProvider.scala:104: Could not find any member to link for "play.api.mvc.Result".
[error]   /**
[error]   ^
[error] /Users/sahebmotiani/github/play-silhouette/silhouette/app/com/mohiva/play/silhouette/impl/providers/SocialProvider.scala:75: Could not find any member to link for "AuthInfo".
[error]   /**
[error]   ^
[error] 8 errors found
[error] (play-silhouette/compile:doc) Scaladoc generation failed
[error] Total time: 36 s, completed 23-Mar-2017 14:37:52
```
@akkie - any idea why this compile:doc sbt task is failing?

akkie:
Sometimes it fails to resolve the classes for the doc task. Sometimes it helps to use the fully qualified class name. For the Silhouette agnostic version we use [sbt-api-mappings](https://github.com/minutemen/silhouette/blob/master/project/plugins.sbt#L23). You could try that but then you must remove the [apiMappings](https://github.com/mohiva/play-silhouette/blob/master/project/BuildSettings.scala#L112). If that not helps you should write the class name in backticks instead of [[]].

akkie:
@Saheb Excellent :+1: Many, many thanks for your time and your patience!

PR #509, opened 2017-04-29
Switch to Play 2.6.0

akkie:


akkie:
Currently we have a problem with changes in the `WSRequest` and `WSResponse` API of Play.  In Play 2.5 we was able to mock `WSRequest` and `WSResponse` with mockito. With Play 2.6 this isn't longer possible because the `WSRequest` and the `WSResponse` traits use now abstract [type members](https://github.com/playframework/playframework/blob/7b96f446cdb328b98121eba664161d09cebd73d2/framework/src/play-ws/src/main/scala/play/api/libs/ws/WSRequest.scala#L15) for the return types and mockito cannot handle this.

The [suggestion](https://groups.google.com/d/msg/play-framework/mYPRPbfa8Uk/Bmkmy40gAwAJ) from @wsargent was to switch to [play-mockws](https://github.com/leanovate/play-mockws). I'm not sure if it's a good idea to switch to a library that as a Play dependency, because then we must depend on the release cycle of this library. If there is no final release for the next Play version of the mocking library, then we cannot release the final version of Silhouette. We had this issues with the seed templates before and that was really annoying.

epot:
Since the 5.0.0 milestone seems 100% complete, when is it planned to be released?

akkie:
In the next few days

epot:
Awesome, looking forward to that :). Thanks!

PR #512, opened 2017-06-03
Adds Cookie encoding to DefaultSocialStateHandler

halfhp:
Disclaimer: I'm new to Scala, to Silhouette and I don't actually have a fully functional app yet so I have no way of knowing if this is the right way to fix this potential bug, let a lone whether or not it creates other bugs.

While trying to get the 2.6 snapshot working with `DefaultSocialStateHandler` and a `CsrfStateItemHandler` I was getting encoding exceptions when Silhouette was attempting to deserialize the csrf cookie post Oauth2 handshake, which lead me to the line above.  Looking at the changelog for 2.6 I noticed that `ItemStructure.asString` was renamed `ItemStructure.asString`.  I'm not sure if this was an accident or not so I opted to submit a fix that attempts to use the new method name.

I've submitted Issue #512 to track this.


halfhp:
Obviously this broke tests.  Will wait for some guidance on what the actual expected behavior is meant to be here, before making additional changes.

akkie:
Fixed width #513

PR #517, opened 2017-06-15
Fix for CookieAuthenticator not working with Play 2.6.0-RC2

sgoyal01:
In play-silhouette with play 2.6.0-RC2, authenticator cookie is not getting returned after successful sign in. This fix removes the default values for copy method in AuthenticatorResult which was leading to cleaning up of authenticator cookie.

Providing default for copy leads to emptying out of parameters in scala. Here is a simple program to understand this behavior in scala:

```scala
case class Foo(p1: String, p2: String) {}
class Bar(p1: String, p2: String) extends Foo(p1, p2) {
  override def copy(arg1: String = "", arg2: String = ""): Foo = { Foo(arg1, arg2) }
}
val x = new Bar("hello", "world")
val y: Foo = x
val z = y.copy(p2 = "blah")
```

Now z is Foo(,blah) instead of Foo("hello", "blah")

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/api/services/AuthenticatorService.scala
@@ -59,9 +59,9 @@ class AuthenticatorResult(result: Result)
   override def copy(
     header: ResponseHeader,
     body: HttpEntity,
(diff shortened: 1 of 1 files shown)


akkie:
Thanks :+1:

PR #519, opened 2017-06-28
Implement BCryptSha256PasswordHasher to avoid password truncating issue

### LINKED ISSUE #504: Create a BCryptSHA256PasswordHasher that supports passwords >= 72 characters
### Expected Behavior

Please describe the expected behavior of the issue, starting from the first action.

1. sign up with password > 72 characters long
2. attempt sign in with a different password with the same 72 character prefex as original password
3. sign in is not successful

### Actual Behavior

1. sign up with password > 72 characters long
2. attempt sign in with a different password with the same 72 character prefex as original password
3. sign in is successful

See the [django docs](https://docs.djangoproject.com/en/1.10/topics/auth/passwords/#using-bcrypt-with-django) for an example of a different framework with the same issue

I propose the the BCryptPasswordHasher raises an exception when presented with a password > 72 characters long, and BCrypt256PasswordHasher is the new default.

Docs should be updated on how to configure both to roll old passwords to new passwords.

akkie:
Fixes #504


graingert:
Super nice, do you have any documentation on how to roll forwards?

PR #523, opened 2017-07-03
Authenticators should keep other request parts in test mode

akkie:
Some authenticators override the request attributes and remove all other request parts.


EnverOsmanov:
I was exploring this issue but have been late for 4 hours :)
Great job akkie.

akkie:
@EnverOsmanov I've released a RC2 with the fix.

PR #533, opened 2017-11-11
Log warning message when validating csrf token item handler state

kurochenko:
## Purpose

When using `DefaultSocialStateHandler` with `CsrfStateItemHandler` it might happen that `CsrfStateItemHandler.canHandle` would return `false` and user would not know exactly whats wrong because `clientState` `Failure` was ignored. 

I've added `warn` log before returning `false` so user can have an idea during development process.

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
@@ -17,6 +17,7 @@ package com.mohiva.play.silhouette.impl.providers.state
 
 import javax.inject.Inject
 
+import com.mohiva.play.silhouette.api.Logger
 import com.mohiva.play.silhouette.api.crypto.Signer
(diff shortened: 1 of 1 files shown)


silhouette/app/com/mohiva/play/silhouette/impl/providers/state/CsrfStateItemHandler.scala
```diff
@@ -108,7 +109,9 @@ class CsrfStateItemHandler @Inject() (
     item.id == ID && {
       clientState match {
         case Success(i) => i == item.data.as[Item]
-        case Failure(_) => false
+        case Failure(e) =>
+          logger.warn(e.getMessage)
```
akkie: Could you please add the exception also to the logger, so that the user can see the stack trace in the log.

kurochenko: Done

akkie:
Thanks :+1:

PR #541, opened 2018-06-07
Added support for setting the SameSite option on cookies.

Asamsig:
# Pull Request Checklist

* [x] Have you added copyright headers to new files?
* [x] Have you suggest documentation edits?

## Purpose

Makes it possible to set the SameSite option on cookies.

## Background Context

~~I chose to use a String representation, since using the Cookie.SameSite type, would break current solutions using Ficus. Alternatively we could create our own Enumeration and convert that to the Play Cookie.SameSite type.~~ EDIT: We're now using the Cookie.SameSite type.

I chose the default to be `Lax`, since from looking at other settings, we want to recommend a sane default, where as `Strict`, might be too harsh, and `None` would simply do nothing.

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/impl/authenticators/CookieAuthenticator.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/authenticators/CookieAuthenticator.scala
@@ -254,7 +254,8 @@ class CookieAuthenticatorService(
         path = settings.cookiePath,
         domain = settings.cookieDomain,
         secure = settings.secureCookie,
-        httpOnly = settings.httpOnlyCookie
+        httpOnly = settings.httpOnlyCookie,
+        sameSite = settings.sameSite
       )
     }.recover {
       case e => throw new AuthenticatorInitializationException(InitError.format(ID, authenticator), e)
@@ -329,7 +330,8 @@ class CookieAuthenticatorService(
         path = settings.cookiePath,
         domain = settings.cookieDomain,
         secure = settings.secureCookie,
-        httpOnly = settings.httpOnlyCookie
+        httpOnly = settings.httpOnlyCookie,
+        sameSite = settings.sameSite
       ))))
     }).recover {
       case e => throw new AuthenticatorUpdateException(UpdateError.format(ID, authenticator), e)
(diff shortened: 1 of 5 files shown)


akkie:
Thanks @Asamsig :+1: 

In my opinion the lib should use the `Cookie.SameSite` type, because the lib itself isn't bound to ficus. Ficus is used by the seed template but this is only one way of wiring config objects. There are plenty of configuration libs out there that can be used for that task and which maybe can use the type provided by Play. For the seed template we can use our own Ficus `ValueReader` to get the type working with Play.

Asamsig:
Okay, I'll make that change 👍

Asamsig:
I've changed to the `Cookie.SameSite` type. You can ping me, when the new version is released, then I can help update the seed template if you'd like.

akkie:
@Asamsig Awesome :+1:  I'll do that.

akkie:
@Asamsig I've just released version 5.0.5

Enalmada:
@Asamsig Lax seems like a smart default and I am disappointed that [play doesn't default to it](https://github.com/playframework/playframework/issues/7316).

nafg:
This is a breaking change. It forces you to define a `ValueReader[Option[Option[Cookie.SameSite]]]`. Otherwise it cannot derive a ValueReader for CookieAuthenticatorSettings.

Enalmada:
Look at the linked ticket in play-silhouette-seed for migration code.  This is what you need put in your SilhouetteModule thanks to @Asamsig:
```
import com.typesafe.config.Config
import net.ceedubs.ficus.readers.ValueReader
import play.api.mvc.{Cookie}

implicit val sameSiteReader: ValueReader[Option[Option[Cookie.SameSite]]] =
  (config: Config, path: String) => {
    if (config.hasPathOrNull(path)) {
      if (config.getIsNull(path))
        Some(None)
      else {
        Some(Cookie.SameSite.parse(config.getString(path)))
      }
    } else {
      None
    }
  }
```

nafg:
Actually this suffices:

```scala
implicit val sameSiteReader: ValueReader[Option[Cookie.SameSite]] =
    ValueReader.relative(cfg => Cookie.SameSite.parse(cfg.as[String]))
```

akkie:
@nafg Could you please create a PR against the seed template with the value reader?

Asamsig:
I tested the provided code from @nafg on my local project, and it doesn't seem to work.

Also keep in mind, the reason the code is quite verbose in the seed project, is because it handles these cases:
Not set, set None, will use default ('Lax')
Set to null, set Some(None), will use 'No Restriction'
Set to a string value try to match, Some(Option(string))

On further consideration, it should probably either log a warning or throw, when given a String value that doesn't match a setting, since otherwise it simply downgrades to 'No Restriction'.

Issue #555, opened 2019-05-13
Support for Google TOTP 2-Factor authentication


bravegag:
Is there support for Google TOTP already? if not, can someone point me to directions where to look to integrate it? by this I mean the silhouette idiomatic way to add such extensions.

I have a reference in another seed: 
https://github.com/bravegag/play-authenticate-usage-scala

that uses: 
https://github.com/wstrange/GoogleAuth

akkie:
This would be a good addition. In Silhouette you would create a new provider. A pull request is welcome.

bravegag:
Hi @akkie, we have the PR ready to merge .. please let us (@dloshkarev) know if there is anything you need from us.
Cheers!

akkie:
@bravegag I'm away from home this week. I'll take a look at it next week.

akkie:
Fixed with #557

PR #557, opened 2019-05-20
TOTP using Google Authenticator

bravegag:
We (@dloshkarev and I) have a PR ready corresponding to issue #555. 

Sample `play-silhouette-seed` project here:
https://github.com/bravegag/play-silhouette-seed/tree/googleauth

I will merge the `googleauth` branch once the PR goes through.


akkie:
At the first glance it looks awesome :+1: Give me some time to dive deeper into the code.

bravegag:
@akkie thank you! we went through a lot of iterations including the correct handling of scratch / recovery codes. Take a look at my Play-Silhouette-Seed as reference for the complete working usage.

Let me know whether you have any questions. We can use the chat if you prefer.

akkie:
Could you please create a suggested edit with documentation about the provider to https://www.silhouette.rocks/docs/providers ?

bravegag:
@akkie sure! Do I have permission to edit?

akkie:
Every user can suggest and edit on readme.io. The PR to the seed would make sense :+1:

bravegag:
@akkie I'm super busy but @dloshkarev will help with the documentation. Are you waiting for the docs before merging this PR? would it be possible to have it before?

bravegag:
Closed by mistake `Ctrl-Return` :(

dloshkarev:
@akkie i've added "suggested edit" - https://www.silhouette.rocks/suggested-edits/5d129572f0954900468ed4d5

akkie:
Correct me if I'm wrong, but it seems that the scratch code is more Google Authenticator specific? The [RFC 6238](https://tools.ietf.org/html/rfc6238) doesn't contain any information about a scratch code. Is this the case, then I think the TotpProvider interface shouldn't contain any information about a scratch code.

bravegag:
@akkie the RFC are not bothered with implementations, that’s true, but is there a TOTP impl without recovery (aka scratch) codes? I think not. We can move the recovery codes logic to the Google provider impl but this means every other TOTP provider will have to copy-paste the recovery codes logic. Doing so would hardly be a good solution design-wise.

Nobody cares wether the impl is 100% compliant with RFCs. Silhouette is not a JDK.

akkie:
In my opinion, if we provide an interface for TOTP then it should be 100% RCF compliant. I see your point, that most implementations use recovery codes. But there can also be implementations that don't use recovery codes. I think both points are valid.

Anyway, I don't see the need for an interface in this implementation. For me, this is a concrete implementation of the GoogleAuth Java lib implementation. Yes, there can be other providers, but for me it would be OK if they have a complete different design. This is a third party extension, which no core code relies on. There will never be two TOTP providers used at the same time. Recovery codes may not be compatible between different providers, so it's also not easy to switch implementations.

At least, I think the provider should live in it's own package because, otherwise every user of Silhouette would fetch an additional dependency that isn't really needed. We do this already for the CAS provider and the framework agnostic version of Silhouette does this as well for all mostly all implementations.

dloshkarev:
@akkie I removed common TotpProvider interface and put all logic into GoogleTotpProvider

bravegag:
@akkie we have put some effort into this please also help us get this PR merged without waiting for nothing.

silhouette/app/com/mohiva/play/silhouette/impl/providers/totp/GoogleTotpProvider.scala
```diff
@@ -0,0 +1,202 @@
...
+   * check each of the previously hashed scratch codes and find the first one that
+   * matches the one entered by the user. The one found is removed from `totpInfo` and returned
+   * for easy client-side bookkeeping.
+   *
+   * @param totpInfo The original TOTP info containing the hashed scratch codes.
+   * @param plainScratchCode The plain scratch code entered by the user.
+   * @return Some tuple consisting of (`PasswordInfo`, `TotpInfo`) if the authentication was successful, None otherwise.
+   */
+  def authenticate(totpInfo: GoogleTotpInfo, plainScratchCode: String): Future[Option[(PasswordInfo, GoogleTotpInfo)]] = Future {
+    Option(totpInfo).flatMap { totpInfo =>
```
akkie: Why is there an option? To handle a null value?

akkie:
The PR seems good to me expect my online comment. It would be great if you could move the code into it's own package as noted in my last comment. As example you could look at the CAS provider.

I'm not home at the weekend. I'll merge the request at the beginning of the next week.

silhouette/app/com/mohiva/play/silhouette/impl/providers/totp/GoogleTotpProvider.scala
dloshkarev: @akkie I guess it's null check. I didn't change it yet, waiting for @bravegag answer. I guess we don't need that, at least you don't have this kind of checks in your code (e.g com.mohiva.play.silhouette.impl.providers.CredentialsProvider#authenticate)

dloshkarev:
@akkie I moved code into own package

bravegag:
>Why is there an option? To handle a null value?
>
Yes Scala’s idiomatic way to handle null values and avoid the ugly boilerplate if-null-then-else ladders.

akkie:
In Scala we don't use null values. Silhouette is a Scala only library, so we mustn't deal with Java code. In my opinion this null type checks only increase the cyclomatic complexity and make the code more unreadable. I've never seen null type checks for parameters in big Scala projects. Passing a null value to a method is a programming error, that should be fixed by the caller.

This is for Java, but in my opinion, it also applies to Scala.
https://dzone.com/articles/why-i-never-null-check-parameters
https://dev.to/scottshipp/better-null-checking-in-java-ngk

bravegag:
@akkie You’re right, old Java habits ...

akkie:
@bravegag @dloshkarev Thanks for your awesome work :+1:

PR #567, opened 2019-08-28
Fix that GoogleTotpProvider ID is can not override.

yu-ishicawa:
# Pull Request Checklist

* [ ] Have you added copyright headers to new files?
* [ ] Have you suggest documentation edits?
* [ ] Have you added tests for any changed functionality?

## Purpose

I want to create a class that inherits GoogleTotpProider and inject id dynamically.

```
class CustomTotpProvider @Inject()(
    override val id: String,
    passwordHasherRegistry: PasswordHasherRegistry
)(implicit ec: ExecutionContext)
    extends GoogleTotpProvider(passwordHasherRegistry) {
  ...
}
```

However, the authenticate method always returns the same ID ("googleTotp").

```
  def authenticate(sharedKey: String, verificationCode: String): Future[Option[LoginInfo]] = {
    Future(
      if (isVerificationCodeValid(sharedKey, verificationCode)) {
        Some(LoginInfo(ID, sharedKey))
                       ^^
    ...
```

### DIFF
--- a/silhouette-totp/src/main/scala/com/mohiva/play/silhouette/impl/providers/GoogleTotpProvider.scala
+++ b/silhouette-totp/src/main/scala/com/mohiva/play/silhouette/impl/providers/GoogleTotpProvider.scala
@@ -122,7 +122,7 @@ class GoogleTotpProvider @Inject() (injectedPasswordHasherRegistry: PasswordHash
   def authenticate(sharedKey: String, verificationCode: String): Future[Option[LoginInfo]] = {
     Future(
       if (isVerificationCodeValid(sharedKey, verificationCode)) {
-        Some(LoginInfo(ID, sharedKey))
+        Some(LoginInfo(id, sharedKey))
       } else {
         logger.debug(VerificationCodeDoesNotMatch)
         None


bravegag:
Hi @yu-ishicawa, I worked in the `GoogleTotpProvider` implementation. This is not intended to be extended from. If you have your own form of TOTP or 2FA then you should make a sibling (replicate the `GoogleTotpProvider` implementation). Unfortunately we were discouraged from creating a common base class for all types of TOTP implementations. See the discussion here:
https://github.com/mohiva/play-silhouette/pull/557#issuecomment-506626712

In other words, it doesn't make sense what you are trying to do. 

The ID you are trying to use exists to uniquely identify the `GoogleTotpProvider` as provider and it's  not intended for overriding, in fact, it should be final in this implementation.

@akkie See what I meant, it is all about UX, we wanted the base `TotpProvider` :)

akkie:
@bravegag I have to disagree with you here. I don't think this is about changing the functionality of the provider. It's about of changing the ID of the provider, for whatever reason.
One use case would be to use providers with different configurations.

From a logical standpoint, I think this PR is valid and fixes an issue. Code in the provider should use the `id` method and not the constant. I think all other providers do this the same way.

yu-ishicawa:
> @bravegag I have to disagree with you here. I don't think this is about changing the functionality of the provider. It's about of changing the ID of the provider, for whatever reason.
> One use case would be to use providers with different configurations.

Exactly. What I wanted to do was simply to use two different IDs, not to change the behavior of GoogleTotpProvider.

I have defined the CustomTotpProvider initialization as follows:

```
  @Provides
  def provideTotpProvider(
      @Named("googleTotp-provider-id") providerId: String,
      passwordHasherRegistry: PasswordHasherRegistry
  ): GoogleTotpProvider = {
    new CustomTotpProvider(providerId, passwordHasherRegistry)
  }
```

A module defines the ID to inject as follows:

```
  @Provides @Named("googleTotp-provider-id")
  def provideGoogleTotpProviderId: String = "googleTotp-a"
```

The other is defined as follows:

```
  @Provides @Named("googleTotp-provider-id")
  def provideGoogleTotpProviderId: String = "googleTotp-b"
```

Thank you for the discussion.

akkie:
@yu-ishicawa Thanks :+1:

akkie:
@yu-ishicawa I've published version 6.1.1 with your fix. Thanks again

PR #568, opened 2019-10-09
Fixed Profile URL as v1 resource is no longer available

ViniciusMiana:
## Fixes

This fix updates the API url to the v2 resource and the parsing of the json response that comes from it.

## Purpose

This PR allows you to user the LinkedIn provider and get the First and Last name of the logged user. Unfortunatelly the email and picture don't come by default anymore and for that reason were left out.

## Background Context

It has been a while since LinkedIn has discontinued its v1 people resource. As a result, the LinkedInProvider was failing to parse the profile of the logged user.

### DIFF
--- a/silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth2/LinkedInProvider.scala
+++ b/silhouette/app/com/mohiva/play/silhouette/impl/providers/oauth2/LinkedInProvider.scala
@@ -89,11 +89,11 @@ class LinkedInProfileParser extends SocialProfileParser[JsValue, CommonSocialPro
    */
   override def parse(json: JsValue, authInfo: OAuth2Info) = Future.successful {
     val userID = (json \ "id").as[String]
-    val firstName = (json \ "firstName").asOpt[String]
-    val lastName = (json \ "lastName").asOpt[String]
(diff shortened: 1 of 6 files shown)


akkie:
@ViniciusMiana Thanks for the update :+1: Would it be possible to fetch the email and picture in an additional API request and merge it with the default data?

This documentation describes the process of getting all data: https://docs.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/sign-in-with-linkedin?context=linkedin/consumer/context

It would also be nice if you can add the links to the new documentation sources and if you could adapt the tests to the new API.

ViniciusMiana:
Thanks @akkie!  I was aware of the necessary calls, however, since it required greater changes, I was not sure it would be welcome. I will do that as soon as I can.

ViniciusMiana:
Hi @akkie, I made the fixes with a few hacks as the more elegant solution would require to change the interface for a SocialProfileParser and others. Let me know what you think and if any other changes are necessary.

akkie:
Hey @ViniciusMiana, I'll see if I can change the interfaces to cover this scenario, in https://github.com/minutemen/silhouette when I update the LinkedIn provider there.

Thanks for your work :+1:

PR #578, opened 2020-05-03
Fixed Auth0ProfileParser to get user id and name

ymotchi:
- Auth0's userinfo API returns sub as user's ID; doesn't have user_id
- It returns name as user's full name
- Also fixed comment of auth0.scope to get user's profile


akkie:
Thank @ymotchi :+1:

